PR #7, opened 2023-11-29
ACS-6432 Package and publish application in zip format.

tpage-alfresco:


Pawel-608 (approved):
lgtm

pom.xml
```diff
@@ -14,6 +14,7 @@
 
   <modules>
     <module>live-ingester</module>
+    <module>distribution</module>
```
krdabrowski, line 17: Here and above there shouldn't be present the `alfresco-hxinsight-connector-` prefix?

distribution/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <name>Alfresco HxInsight Connector :: Distribution zip</name>
+    <artifactId>alfresco-hxinsight-connector-distribution</artifactId>
+    <packaging>jar</packaging>
+
+    <profiles>
+        <profile>
+            <id>distribution</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
```
krdabrowski, line 17: This is intentional - zip should be created every time app will be built?

mpichura (approved):
Looks Good, Me Too ;)

distribution/pom.xml
krdabrowski: Isn't the distribution module also built every time we use `mvn package` in root folder?
That was my first assumption, that zip file will be created if particular profile will be passed through `-P`

tpage-alfresco: I guess it boils down to expectations about what should happen if you build in the root folder. I would normally expect everything to build by default, but I'm aware this doesn't happen in many projects.

I don't mind adding profiles to support different use cases, but maybe we'll know better what we want after working on the project for a while? I'm happy if you want to make a PR for it or raise a ticket to do something about this?

krdabrowski: I think that we should build distribution `zip` only when it's required, similarly to building Docker images in other projects. This could save us few seconds while building app localy.

krdabrowski: I'm OK with this approach to work a bit on the project and see later if this build of `zip` distribution will give us some gain

PR #8, opened 2023-11-30
ASC-6375: log message if received event of type updated

Pawel-608:


live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/messaging/EventProcessor.java
```diff
@@ -49,10 +51,14 @@ public void process(Exchange exchange)
     {
         RepoEvent<?> event = eventFrom(exchange);
 
-        if (event.getType().equals(EventType.NODE_CREATED.getType()))
+        if (NODE_CREATED.getType().equals(event.getType()))
```
tpage-alfresco, line 54: We can use equality for enum values:
```suggestion
        if (event.getType() == NODE_CREATED.getType())
```

tpage-alfresco (approved):
Yep - this was definitely smaller than we sized it :joy:

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/messaging/EventProcessor.java
Pawel-608: for some reason event.getType() is string not enum😂

PR #10, opened 2023-12-01
ACS-6432 Add docker-compose.

tpage-alfresco:
To get started with this:
```
mvn clean install
./scripts/ci/buildDockerImages.sh
cd distribution/src/main/resources/docker-compose
docker compose -f docker-compose.yml --project-name dev up
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ target
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 
 hs_err_pid*
+
+# This file is generated from distribution/src/main/resources/docker-compose-origin/.env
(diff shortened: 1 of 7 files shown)


Pawel-608 (approved):
LGTM, just pmd check is failing for some reason

krdabrowski (approved):
LGTM

tpage-alfresco:
@mpichura We don't "need" ESC in the d-c, but since this is going to be used by us then I think it will make the environment more pleasant to interact with. I don't think we'll use this for automated testing.

tpage-alfresco:
PMD build was failing because it wasn't possible to build the old code using the new command.

PR #11, opened 2023-12-04
ASC-6376: add sns integration

Pawel-608:


live-ingester/src/main/resources/application.yml
```diff
@@ -0,0 +1,15 @@
...
+      overrideEndpoint: true
+      uriEndpointOverride: http://localhost:4566
+      message-group-id-strategy: useExchangeId
+      secretKey: dummy_secret_key
+      accessKey: dummy_access_key
+
+alfresco:
+  ingester:
+    messaging:
+        in.endpoint: activemq:queue:org.alfresco.ingester.event
```
tpage-alfresco, line 14: I changed this queue name so that the connection worked ootb.

Pawel-608:
I checked that and it look like everything is ok for now because we do not send anything to sns yet so I would merge this pr and add localstack to distribution docer-compose so we can connect to sns

tpage-alfresco (approved):
Ok - LGTM then!

PR #14, opened 2023-12-05
ASC-6376: add localstack to distribution docker compose

Pawel-608:


distribution/src/main/resources/docker-compose/docker-compose.yml
```diff
@@ -179,6 +179,14 @@ services:
       - activemq
     environment:
       SPRING_ACTIVEMQ_BROKERURL: nio://activemq:61616
+  localstack:
+    image: localstack/localstack:${LOCALSTACK_TAG}
```
tpage-alfresco, line 183: You need to update the `.env` template in order for this to link to the maven property.

PR #16, opened 2023-12-06
ASC-6376: set new node metadata properties

Pawel-608:


live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/model/out/event/UpdateNodeMetadataEvent.java
```diff
@@ -0,0 +1,60 @@
...
+
+import lombok.NoArgsConstructor;
+import org.alfresco.hxi_connector.live_ingester.domain.model.out.NodeProperty;
+
+@NoArgsConstructor(staticName = "create")
+public class UpdateNodeMetadataEvent
+{
+
+    @JsonProperty("setProperties")
+    private final List<NodeProperty<?>> metadataPropertiesToSet = new ArrayList<>();
```
tpage-alfresco, line 42: I think this should be more like a `Map<String, Object>` since we can set each property to one value.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/model/in/IngestNewNodeEvent.java
```diff
@@ -0,0 +1,33 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+
+package org.alfresco.hxi_connector.live_ingester.domain.model.in;
+
+public record IngestNewNodeEvent(
```
tpage-alfresco, line 29: Small point of confusion for me - the class name suggests this is only for new nodes and the PR message suggests this is for updates.

tpage-alfresco:
Nice work @Pawel-608

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/model/out/event/UpdateNodeMetadataEvent.java
Pawel-608: yep, you're right, I'll change it to Map<String, NodeProperty >

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/model/out/PredefinedNodeProperty.java
```diff
@@ -0,0 +1,52 @@
...
+{
+    public static final PredefinedNodeProperty<String> NAME = new PredefinedNodeProperty<>("name");
+    public static final PredefinedNodeProperty<String> PRIMARY_ASSOC_Q_NAME = new PredefinedNodeProperty<>("primaryAssocQName");
+    public static final PredefinedNodeProperty<String> TYPE = new PredefinedNodeProperty<>("type");
+    public static final PredefinedNodeProperty<String> CREATED_BY_USER_WITH_ID = new PredefinedNodeProperty<>("createdByUserWithId");
+    public static final PredefinedNodeProperty<String> MODIFIED_BY_USER_WITH_ID = new PredefinedNodeProperty<>("modifiedByUserWithId");
+    public static final PredefinedNodeProperty<Set<String>> ASPECTS_NAMES = new PredefinedNodeProperty<>("aspectsNames");
+    public static final PredefinedNodeProperty<Boolean> IS_FOLDER = new PredefinedNodeProperty<>("isFolder");
+    public static final PredefinedNodeProperty<Boolean> IS_FILE = new PredefinedNodeProperty<>("isFile");
+    public static final PredefinedNodeProperty<Long> CREATED_AT = new PredefinedNodeProperty<>("createdAt");
```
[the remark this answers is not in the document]
Pawel-608: check added

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/model/in/IngestNewNodeEvent.java
Pawel-608: ah ok I can see it now, I'll change pr name

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/messaging/config/jackson/UpdateNodeMetadataEventSerializer.java
```diff
@@ -0,0 +1,102 @@
...
+            jgen.writeObjectField(property.name(), property.value());
+            jgen.writeEndObject();
+        }
+        catch (IOException e)
+        {
+            throw new JsonSerializationException("Property serialization failed", e);
+        }
+    }
+
+    private void writeProperty(JsonGenerator jgen, String property)
```
tpage-alfresco, line 91: Maybe could rename this to:
```
    private void writePropertyName(JsonGenerator jgen, String propertyName)
```

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/messaging/config/jackson/UpdateNodeMetadataEventSerializerTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Test
+    public void shouldSerializePropertiesToSet()
+    {
+        UpdateNodeMetadataEvent event = UpdateNodeMetadataEvent.create()
+                                            .set(NAME.withValue("some-name"))
+                                            .set(IS_FILE.withValue(true))
+                                            .set(MODIFIED_BY_USER_WITH_ID.withValue("000-000-000"));
+
+        String expectedJson = getExpectedMapping("with-properties-to-set.json");
```
tpage-alfresco, line 66: Now we can do multi-line strings in Java then I would inline these expected results. It's a pain to try to find out what the test does if you have to switch between several different files and these files are only a few lines each.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/messaging/config/jackson/UpdateNodeMetadataEventSerializerTest.java
```diff
@@ -0,0 +1,103 @@
...
+        String actualJson = serialize(event);
+
+        assertEquals(expectedJson, actualJson);
```
tpage-alfresco, line 69: Isn't this non-deterministic? Do we somehow guarantee the order of the properties?

Pawel-608: it is probably not guaranteed but it works ;) so for now I would leave it as it is as it is the simplest way to test it and in case of any problems we can always find some better way

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/messaging/config/jackson/UpdateNodeMetadataEventSerializerTest.java
Pawel-608: I was considering that and tbh it is probably better way. I'll change it

Pawel-608: done

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/util/TestUtils.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package org.alfresco.hxi_connector.live_ingester.util;
+
+import static java.util.function.Predicate.not;
+
+import java.util.Collection;
+import java.util.List;
+
+public class TestUtils
```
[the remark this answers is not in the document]
Pawel-608: fixed

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/event/IngestNewNodeEventHandlerTest.java
```diff
@@ -0,0 +1,114 @@
...
+    private static final String NODE_PRIMARY_ASSOC_Q_NAME = "cm:test-name";
+    private static final String NODE_TYPE = "cm:folder";
+    private static final String NODE_CREATED_BY_USER_WITH_ID = "admin";
+    private static final String NODE_MODIFIED_BY_USER_WITH_ID = "admin";
+    private static final Set<String> NODE_ASPECT_NAMES = Set.of(
+        "cm:titled",
+        "cm:auditable");
+    private static final boolean NODE_IS_FOLDER = true;
+    private static final boolean NODE_IS_FILE = false;
+    private static final long NODE_CREATED_AT = 1690000000050L;
```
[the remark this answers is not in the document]
Pawel-608: fixed

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/event/IngestNewNodeEventHandlerTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class IngestNewNodeEventHandlerTest
+{
+    private static final long EVENT_TIMESTAMP = 1690000000100L;
```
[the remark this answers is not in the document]
Pawel-608: fixed

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/messaging/in/RepoEventMapperTest.java
```diff
@@ -0,0 +1,196 @@
...
+    private static final boolean NODE_IS_FOLDER = true;
+    private static final boolean NODE_IS_FILE = false;
+    private static final long NODE_CREATED_AT = 1_690_000_000_050L;
+    private static final NodeProperty<String> NODE_TITLE = new NodeProperty<>("cm:title", "some title");
+    private static final Set<NodeProperty<?>> NODE_PROPERTIES = Set.of(NODE_TITLE);
+    private final CamelEventMapper camelEventMapper = new CamelEventMapper(ObjectMapperFactory.createInstance());
+    private final RepoEventMapper repoEventMapper = new RepoEventMapper();
+
+    @Test
+    void mapToIngestNewNodeEvent()
```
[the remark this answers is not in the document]
tpage-alfresco: I'm not hugely keen on either. Looking at the method being tested then the main part that has logic is this:
```
node.getProperties()
                    .entrySet()
                    .stream()
                    .filter(property -> Objects.nonNull(property.getKey()) && Objects.nonNull(property.getValue()))
                    .map(property -> new NodeProperty<>(property.getKey(), property.getValue()))
                    .collect(Collectors.toSet()))
```
How about if we split this out into a separate class and write a mock based unit test for it, and then the mock-based test for this class would be much simpler.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/messaging/in/RepoEventMapperTest.java
```diff
@@ -0,0 +1,196 @@
...
+        when(nodeResource.getName()).thenReturn(NODE_NAME);
+        when(nodeResource.getPrimaryAssocQName()).thenReturn(NODE_PRIMARY_ASSOC_Q_NAME);
+        when(nodeResource.getNodeType()).thenReturn(NODE_TYPE);
+        when(nodeResource.getCreatedByUser()).thenReturn(nodeCreator);
+        when(nodeResource.getModifiedByUser()).thenReturn(nodeModifier);
+        when(nodeResource.getAspectNames()).thenReturn(NODE_ASPECT_NAMES);
+        when(nodeResource.isFile()).thenReturn(NODE_IS_FILE);
+        when(nodeResource.isFolder()).thenReturn(NODE_IS_FOLDER);
+        when(nodeResource.getCreatedAt()).thenReturn(dateFromTimestamp(NODE_CREATED_AT));
+        when(nodeResource.getProperties()).thenReturn(createPropertiesMap("cm:title", "some title", "cm:description", null));
```
tpage-alfresco, line 134: Could we simplify this by returning a real NodeResource from the mock?

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/messaging/in/RepoEventMapperTest.java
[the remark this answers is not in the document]
Pawel-608: I don't think it is good approach as this class is already really small and does almost nothing, also we can test this piece really easily by passing node with just two properties: "sometthing": "something" and "some_other_thing": null to check if null is filtered out. Btw this filter is also checking if key is not null - do you think key could be null?

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/messaging/in/RepoEventMapperTest.java
Pawel-608: yep, we can actually do that - should I remove the other test that reads json and use this approach?

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/messaging/in/RepoEventMapperTest.java
```diff
@@ -0,0 +1,196 @@
...
+            NODE_TYPE,
+            NODE_CREATED_BY_USER_WITH_ID,
+            NODE_MODIFIED_BY_USER_WITH_ID,
+            NODE_ASPECT_NAMES,
+            NODE_IS_FILE,
+            NODE_IS_FOLDER,
+            NODE_CREATED_AT,
+            NODE_PROPERTIES);
+
+        IngestNewNodeEvent expectedEvent = new IngestNewNodeEvent(
```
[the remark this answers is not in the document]
tpage-alfresco: Yep - I meant to move it to the "then" ("after the when"). Let's try this test using a real NodeResource.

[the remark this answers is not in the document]
Pawel-608: ahh, I missed it, done

tpage-alfresco (approved):
LGTM

PR #18, opened 2023-12-12
ACS-6479: apply new code format

Pawel-608:


README.md
```diff
@@ -14,7 +14,6 @@ To reformat files you can use:
 ```bash
 mvn spotless:apply
 ```
-or load `alfresco-formatter.xml` to IntelliJ
```
tpage-alfresco, line 17: Is this no good?

tpage-alfresco (approved):
LGTM

README.md
Pawel-608: I felt it would be better to move info about formatter as close to the `Code Quality` header as possible since most people would like to just import it without having to know anything about spotless

PR #20, opened 2023-12-12
ACS-6379: [HxI LI] S3 Credentials - PART I

krdabrowski:
This PR only adds part of the functionality:
- adding component allowing to upload objects to S3 storage using preSigned URL
- adding integration test based on Testcontainers and LocalStack
- fixed Maven and test config allowing integration tests to run using properties from pom file

what is still missing and will be added in next part:
- way to call HxI API, which will return preSigned URL

For more see: https://alfresco.atlassian.net/browse/ACS-6379


live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.testcontainers.utility.DockerImageName;
+
+import org.alfresco.hxi_connector.live_ingester.storage.local.LocalStorageClient;
+import org.alfresco.hxi_connector.live_ingester.storage.local.LocalStorageConfig;
+import org.alfresco.hxi_connector.live_ingester.utils.DockerTags;
+
+@SpringBootTest(classes = {LocalStorageConfig.class, SignedStorageClient.class})
+@ActiveProfiles({"test"})
+@Testcontainers
+class StorageClientIntegrationTest
```
Pawel-608, line 56: I wonder if we could separate unit and integration tests by adding a new test source, like here: https://github.com/Alfresco/hxi-connector/tree/feature/add-it-test. It is really easy to configure and can be really handy as you don't mix unrelated stuff. wdyt @krdabrowski @tpage-alfresco @mpichura?

krdabrowski: I was also thinking about that and so far, I don't see any contraindications 🤔 I can adjust it, but let see what rest will say

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/utils/DockerTags.java
```diff
@@ -0,0 +1,87 @@
...
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import lombok.SneakyThrows;
+
+/**
+ * Allows to access properties from 'docker-tags.properties' file. It's required to run: `mvn package` before accessing the properties file, to allow Maven filter out and replace variables.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class DockerTags
```
Pawel-608, line 41: There are two test utils (`utils` and `util`) packages now ;) (probably because of rebase) I would move everything to one of them and remove other one

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/utils/DockerTags.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        return Optional.ofNullable(properties).map(p -> p.getProperty(key)).orElse(null);
+    }
+
+    public static String getOrDefault(String propertyKey, String defaultValue)
+    {
+        if (properties == null)
+        {
+            loadProperties();
```
Pawel-608, line 62: I am wondering if maybe we could use just `getProperty()` that will throw exception if property is not found and also make `loadProperties()` throw an error if properties doesn't exists as we probably expect that it should always be available and this approach will allow us to fail fast

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
```diff
@@ -0,0 +1,113 @@
...
+    private StorageClient storageClient;
+
+    @BeforeAll
+    static void beforeAll() throws IOException, InterruptedException
+    {
+        localStack.execInContainer("awslocal", "s3api", "create-bucket", "--bucket", BUCKET_NAME);
+    }
+
+    @Test
+    void testUpload() throws IOException
```
Pawel-608, line 81: maybe it is worth to add @SneakyThrows?

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
```diff
@@ -0,0 +1,113 @@
...
+    {
+        List<String> freshBucketContent = testS3Storage.listBucketContent(BUCKET_NAME);
+        assertThat(freshBucketContent)
+                .isNotNull()
+                .isEmpty();
+
+        // given
+        try (InputStream fileInputStream = new ByteArrayInputStream(OBJECT_CONTENT.getBytes()))
```
Pawel-608, line 89: I am not sure as I never used it but when I was checking lombok features I found this annotation https://projectlombok.org/features/Cleanup, dyt we can use it here?

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
tpage-alfresco: Separating `src/test/java` from `src/integration-test/java` is fine. Doing a quick scan of our existing projects then it seems more standard to rename the other way - into `src/unit-test/java` and `src/test/java`.

distribution/pom.xml
```diff
@@ -73,7 +73,7 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-assembly-plugin</artifactId>
-                        <version>3.6.0</version>
+                        <version>${maven-assembly-plugin.version}</version>
```
tpage-alfresco, line 76: I guess possibly even better is to add a pluginManagement section if we want to manage this centrally.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClient.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.alfresco.hxi_connector.live_ingester.storage;
+
+import java.io.File;
+import java.io.InputStream;
+import java.net.URL;
+
+import org.apache.http.StatusLine;
+
+public interface StorageClient
```
Pawel-608, line 34: I think we should move this interface to domain and probably adjust it a little as it shows internal implementation details (use of apache and pre-signed urls) but it is probably something we can do later (I have no idea how we want to use it so idk how we can improve it😂)

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/storage/SignedStorageClient.java
```diff
@@ -0,0 +1,88 @@
...
+            httpPut.setEntity(entity);
+            httpPut.setHeader(CONTENT_TYPE, contentType);
+            // TODO If additional metadata will be required uncomment bellow, otherwise removed it
+            // metadata.forEach((k, v) -> httpPut.setHeader("x-amz-meta-" + k, v));
+
+            HttpResponse response = httpClient.execute(httpPut);
+            return response.getStatusLine();
+        } catch (URISyntaxException e)
+        {
+            throw new IllegalArgumentException("Pre Signed URL cannot be parsed to URI", e);
```
Pawel-608, line 82: I would stick to `LiveIngesterRuntimeException` and its subclasses (you can always create custom one)

live-ingester/pom.xml
```diff
@@ -53,6 +53,22 @@
 			<artifactId>acs-event-model</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>org.testcontainers</groupId>
+			<artifactId>junit-jupiter</artifactId>
+			<scope>test</scope>
```
tpage-alfresco, line 59: It's probably fairly safe to put the scope of these three dependencies in the `dependencyManagement` section alongside the version.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/storage/SignedStorageClient.java
```diff
@@ -0,0 +1,88 @@
...
+        try (CloseableHttpClient httpClient = HttpClients.createDefault())
+        {
+            HttpPut httpPut = new HttpPut(preSignedUrl.toURI());
+            HttpEntity entity = EntityBuilder.create()
+                    .setStream(inputStream)
+                    .build();
+            httpPut.setEntity(entity);
+            httpPut.setHeader(CONTENT_TYPE, contentType);
+            // TODO If additional metadata will be required uncomment bellow, otherwise removed it
+            // metadata.forEach((k, v) -> httpPut.setHeader("x-amz-meta-" + k, v));
```
Pawel-608, line 76: I would delete this comment

Pawel-608 (approved):
generally LGTM, the only thing is that we probably will need to adjust `StorageClient` interface at some point

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
```diff
@@ -0,0 +1,113 @@
...
+    private static final String LOCALSTACK_IMAGE = "localstack/localstack";
+    private static final String LOCALSTACK_TAG = DockerTags.getOrDefault("localstack.tag", "3.0.2");
+    private static final String BUCKET_NAME = "test-hxinsight-bucket";
+    private static final String OBJECT_KEY = "dummy.txt";
+    private static final String OBJECT_CONTENT = "dummy's file dummy content";
+    private static final String OBJECT_CONTENT_TYPE = "plain/text";
+
+    @Container
+    @SuppressWarnings("PMD.FieldNamingConventions")
+    private static final LocalStackContainer localStack = new LocalStackContainer(DockerImageName.parse(LOCALSTACK_IMAGE).withTag(LOCALSTACK_TAG));
```
tpage-alfresco, line 67: Why didn't you want to call this `LOCAL_STACK`? Maybe you could add a comment to explain the suppression if this has to match with somewhere else in the code?

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
Pawel-608: can we separate it then?

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
```diff
@@ -0,0 +1,113 @@
...
+    {
+        localStack.execInContainer("awslocal", "s3api", "create-bucket", "--bucket", BUCKET_NAME);
+    }
+
+    @Test
+    void testUpload() throws IOException
+    {
+        List<String> freshBucketContent = testS3Storage.listBucketContent(BUCKET_NAME);
+        assertThat(freshBucketContent)
+                .isNotNull()
```
tpage-alfresco, line 85: I don't think you need this check - `isEmpty` already covers it.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+            // when
+            int actualStatusCode = storageClient.upload(fileInputStream, OBJECT_CONTENT_TYPE, preSignedUrl).getStatusCode();
+
+            // then
+            assertThat(actualStatusCode).isEqualTo(200);
+            List<String> actualBucketContent = testS3Storage.listBucketContent(BUCKET_NAME);
+            assertThat(actualBucketContent)
+                    .isNotNull()
+                    .containsOnly(OBJECT_KEY);
```
tpage-alfresco, line 101: I don't think you need the `isNotNull` here, and I think it would be more accurate to replace `containsOnly` with `isEqualTo` (since otherwise `[OBJECT_KEY, OBJECT_KEY]` will pass the test).

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
mpichura: It's definitely worth separating out different types of tests.
Although I usually used to do that with packages (unit tests in same packages as tested classes, integration tests in separate package), I am okay with separate test sources as this at times be even more effective.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/utils/DockerTags.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        String value = defaultValue;
+        if (properties != null)
+        {
+            String property = properties.getProperty(propertyKey);
+            if (property != null && !property.startsWith("@") && !property.endsWith("@"))
```
tpage-alfresco, line 69: I'm not sure I understand the idea behind the second part of this assertion - if a property is not substituted properly by maven then we use the default value? I think this would lead to hard to track down issues - wouldn't it be easier if we just got the `@something@` value instead and then saw this mentioned in the resulting error message?

tpage-alfresco (commented):
Overall this looks great! I tried out the StorageClientIntegrationTest locally and it was really fast. Thanks Krys!

distribution/pom.xml
krdabrowski: I will do it with pleasure

live-ingester/pom.xml
krdabrowski: In the other hand in future it still will be possible to to remove the `test` scope from `dependencyManagement` section

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/utils/DockerTags.java
krdabrowski: Didn't notice that - I will fix it, thanks

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
krdabrowski: Over test methods it doesn't bring any value IMO

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
krdabrowski: Ok, I will work on that - I will go with `src/test/java` and `src/integration-test/java` as it's more popular convention in Java world. Also I think, that unit test are these primary kind of tests and they shouldn't be offended with some prefixes/suffixes.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClient.java
krdabrowski: So this the issue at this stage - I'm not sure how it will look finally. It will clarify in the second part of this task. In general it's a representation of an service/component, so I'm not sure if `domain` is the right place. Overall I would leave it as it is, till the second part is done.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/storage/SignedStorageClient.java
krdabrowski: I'm not sure if it won't be necessary in the second part, for now I would leave it.

Pawel-608: Are you sure you won't forget to remove it if you don't need it?😂

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
krdabrowski: Yeah, so there is a live debate in Java society about what should be written using SNAKE_CASE an what not. I believe that only constants should be written this way. What is a constant is very debatable, there is quite cool explanation here: https://stackoverflow.com/a/12069687/2211974 - I like it 😅 Overall IMO `localStack` is a final reference not a CONSTANT, so I called it using camelCase.

tpage-alfresco: Ah - yes, I read that before. However since we've all had years of seeing `CAPS == static final` then the discussion seemed a bit moot. I think the convenience of using the `@Slf4j` annotation is worth an exception however. Overall I don't mind either way, but if we want to use this style more widely we should probably remove the PMD check.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/utils/DockerTags.java
krdabrowski: This is true, but I must admit, that it's quite annoying when you open a project, try to run a test and it fails because you need to make some step (e.g. `mvn package`) earlier. I like tests which run smoothly without any prerequisites. This is the reason why there are two methods: `getProperty()` and `getOrDefault()`, first one in did should fail if file is missing, but the second one shouldn't care if the file or property exists - if not return default value.
In the other hand there is this small risk, that even on CI something could go wrong with Docker image tag substitution because of what default value would be picked up and we wouldn't notice it. At the same time it's hard for me to image a situation where Maven would not handle the resource filtering correctly while building the app - something would have to be really messed up in config... and even if, would it be so bad to use the default value instead (even outdated)? We need to choose between convenience and reliability - I like when something works straight away, but in this case - having only one prerequisite - I could live with that. Tell me guys @Pawel-608 @tpage-alfresco @mpichura WDYT

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/storage/SignedStorageClient.java
krdabrowski: I'm sure 😅

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
krdabrowski: After small investigation it turned out, that it's not a trivial stuff to do - Maven is designed this way, that all test should be placed in one folder - of course there exist a way to workaround it, but not without issues. In general the community recommendations are: leave all kinds of tests in one (dedicated by Maven) folder, or create a separate Maven module for that: e.g. https://stackoverflow.com/questions/37998040/where-should-the-integration-tests-be-stored-when-using-the-maven-failsafe-plugi or https://stackoverflow.com/questions/46831999/maven-add-integration-tests and plenty more...
I think this issue deserves a separate Jira task, so won't be solved here. FYI @Pawel-608 @tpage-alfresco @mpichura

tpage-alfresco: Definitely fine to raise a separate ticket for this to be done. I found several projects with `src/unit-test/java` so I think we can probably copy something from one of those to set the configuration.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/utils/DockerTags.java
tpage-alfresco: As we discussed we could:
* Do what you've done for all our "real" integration tests
* Add one test that only runs on CI which checks there are no `@something@` properties

We want to add a dev build profile anyway as there are some things that take a while and we don't need to do for each local build (e.g. download the license texts).

krdabrowski:
@tpage-alfresco @Pawel-608 I've updated most of the stuff you mentioned.
The test mentioned here: https://github.com/Alfresco/hxi-connector/pull/20#discussion_r1425627172 I will add in PART II

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/storage/StorageClientIntegrationTest.java
krdabrowski: Task created: https://alfresco.atlassian.net/browse/ACS-6514

PR #21, opened 2023-12-12
ACS-6472 Extract node event mapping.

tpage-alfresco:
This is in preparation for adding an additional mapping to a transform request.

### DIFF
--- a/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/event/UpdateNodeEventMapper.java
+++ b/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/event/UpdateNodeEventMapper.java
@@ -1,0 +1,64 @@
+/*
+ * #%L
+ * Alfresco HX Insight Connector
+ * %%
+ * Copyright (C) 2023 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+
+package org.alfresco.hxi_connector.live_ingester.domain.event;
+
+import static org.alfresco.hxi_connector.live_ingester.domain.model.out.PredefinedNodeProperty.ASPECTS_NAMES;
+import static org.alfresco.hxi_connector.live_ingester.domain.model.out.PredefinedNodeProperty.CREATED_AT;
+import static org.alfresco.hxi_connector.live_ingester.domain.model.out.PredefinedNodeProperty.CREATED_BY_USER_WITH_ID;
+import static org.alfresco.hxi_connector.live_ingester.domain.model.out.PredefinedNodeProperty.IS_FILE;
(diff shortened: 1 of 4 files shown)


Pawel-608 (approved):
LGTM, I would just like to talk about unit testing approach (should we prefer class-level mocks even for data-classes (IngestNewNodeEvent, UpdateNodeMetadataEvent) or maybe use class-level mocks just for classes that are doing some work (EventPublisher, IngestNewNodeHandler) and keep data-classes in test method)

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/event/IngestNewNodeEventHandlerTest.java
```diff
@@ -26,91 +26,43 @@
...
-        ingestNewNodeEventHandler.handle(event);
+        ingestNewNodeEventHandler.handle(ingestNewNodeEvent);
 
         // then
-        verify(eventPublisher).publishMessage(updateNodeMetadataEventCaptor.capture());
-        UpdateNodeMetadataEvent updateNodeMetadataEvent = updateNodeMetadataEventCaptor.getValue();
-
-        assertContainsSameElements(expectedNodePropertiesToSet, updateNodeMetadataEvent.getMetadataPropertiesToSet().values());
-        assertTrue(updateNodeMetadataEvent.getMetadataPropertiesToUnset().isEmpty(), "There should be no properties to unset");
+        verify(eventPublisher).publishMessage(updateNodeMetadataEvent);
```
krdabrowski, line 66: Here also IMO we should choose between Mockito's `when/verify` or `given/then`

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/event/UpdateNodeEventMapper.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.stereotype.Component;
+
+import org.alfresco.hxi_connector.live_ingester.domain.model.in.IngestNewNodeEvent;
+import org.alfresco.hxi_connector.live_ingester.domain.model.in.Node;
+import org.alfresco.hxi_connector.live_ingester.domain.model.out.event.UpdateNodeMetadataEvent;
+
+@Component
+public class UpdateNodeEventMapper
+{
+    public UpdateNodeMetadataEvent map(IngestNewNodeEvent event)
```
krdabrowski, line 48: Somewhere there exists this project: https://mapstruct.org/ which help with mappers generation - maybe it could be useful also for us. In the other hand I'm wondering if it's not an overkill going this way 🤔  WDYT @tpage-alfresco @mpichura @Pawel-608

Pawel-608: I never used it so it is hard to tell if it will work or not, but maybe you could prepare some poc or something? Would we be able to replace this method with it?

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/event/IngestNewNodeEventHandlerTest.java
tpage-alfresco: This is fixed in the next PR: #19

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/event/UpdateNodeEventMapper.java
tpage-alfresco: I'd definitely be interested in seeing an example too. For this example I don't think it would work very well because `UpdateNodeMetadataEvent` is very unstructured, but I didn't look through all the docs on it so I don't know for sure.

tpage-alfresco:
> LGTM, I would just like to talk about unit testing approach (should we prefer class-level mocks even for data-classes (IngestNewNodeEvent, UpdateNodeMetadataEvent) or maybe use class-level mocks just for classes that are doing some work (EventPublisher, IngestNewNodeHandler) and keep data-classes in test method)

This is addressed in the next PR #19

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/event/UpdateNodeEventMapper.java
krdabrowski: Ok, I will try to prepare some PoC tomorrow

PR #22, opened 2023-12-13
Update import ordering.

tpage-alfresco:
Try to change this to be nearer what we have elsewhere. The important line in the change is:
```
<order>\#java|\#javax|\#jakarta,\#,\#org.alfresco,java|javax|jakarta,,org.alfresco</order>
```

This can be replicated in IntelliJ in `Editor/Code Style/Java/Import Layout` by setting "Layout static imports separately" as checked and the order:
```
import static java.*
import static javax.*
import static jakarta.*
<blank line>
import static all other imports
<blank line>
import static org.alfresco.*
<blank line>
import java.*
import javax.*
import jakarta.*
<blank line>
import all other imports
<blank line>
import org.alfresco.*
```

I think in some other projects we're treating `com.*` as a special case, and `org.*` is lumped together with `org.alfresco.*`, but this seems to make more sense to me.  Happy to hear other views on this though!


pom.xml
```diff
@@ -300,7 +300,7 @@
             </eclipse>
 
             <importOrder>
-              <order>\#,\#org.alfresco,java,javax,jakarta,com,lombok,org,org.alfresco</order>
+              <order>\#java|\#javax,\#,\#org.alfresco,java|javax,,org.alfresco</order>
```
krdabrowski, line 303: `jakarta` is not a modern way to say `javax`? Maybe it should be located right after `javax`?

tpage-alfresco: Great suggestion - I'll update.

krdabrowski (approved):

krdabrowski:
Very nice job!

PR #23, opened 2023-12-13
ACS-6376: add support for updating node metadata properties

Pawel-608:


live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/event/UpdateNodeEventMapper.java
```diff
@@ -61,4 +68,50 @@ public UpdateNodeMetadataEvent map(IngestNewNodeEvent event)
...
     }
+
+    public UpdateNodeMetadataEvent map(UpdateNodeEvent event)
+    {
+        UpdateNodeMetadataEvent updateMetadataEvent = UpdateNodeMetadataEvent.create();
+
+        applyFieldDelta(updateMetadataEvent, NAME, event.name());
+        applyFieldDelta(updateMetadataEvent, PRIMARY_ASSOC_Q_NAME, event.primaryAssocQName());
+        applyFieldDelta(updateMetadataEvent, TYPE, event.nodeType());
+        applyFieldDelta(updateMetadataEvent, MODIFIED_BY_USER_WITH_ID, event.modifiedByUserWithId());
```
tpage-alfresco, line 79: Nice thought - the "created" properties shouldn't be able to change. It might be worth testing this in case it's actually possible to retroactively set them.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/event/UpdateNodeEventHandler.java
```diff
@@ -0,0 +1,50 @@
...
+{
+
+    private final UpdateNodeEventMapper updateNodeEventMapper;
+    private final EventPublisher eventPublisher;
+
+    public void handle(UpdateNodeEvent event)
+    {
+        UpdateNodeMetadataEvent updateMetadataEvent = updateNodeEventMapper.map(event);
+
+        eventPublisher.publishMessage(updateMetadataEvent);
```
tpage-alfresco, line 48: We also need to call the `TransformRequester` if the content has been updated. It's fine to handle this in a separate PR though if you want to keep them focused.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/messaging/in/mapper/RepoEventMapper.java
```diff
@@ -67,11 +69,63 @@ public IngestNewNodeEvent mapToIngestNewNodeEvent(RepoEvent<DataAttributes<NodeR
...
+        return new UpdateNodeEvent(
+                event.getTime().toInstant().toEpochMilli(),
+                calculateFieldDelta(event, NodeResource::getName),
+                calculateFieldDelta(event, NodeResource::getPrimaryAssocQName),
+                calculateFieldDelta(event, NodeResource::getNodeType),
+                calculateFieldDelta(event, this::getModifiedByUserWithId),
+                calculateFieldDelta(event, NodeResource::getAspectNames),
+                calculateFieldDelta(event, NodeResource::isFile),
+                calculateFieldDelta(event, NodeResource::isFolder),
+                calculateFieldDelta(event, this::getNodeCustomProperties));
```
tpage-alfresco, line 86: This is all really neat - I like it. I think that you might need to treat the custom properties differently as I think the delta works differently for this one field. For example
```
    "resourceBefore": {
      "@type": "NodeResource",
      "modifiedAt": "2021-01-21T11:14:25.223Z",
      "properties": {
        "cm:title": null,
        "cm:taggable": null,
        "cm:description": null
      },
      "aspectNames": [
        "cm:versionable",
        "cm:author",
        "cm:thumbnailModification",
        "cm:titled",
        "rn:renditioned",
        "cm:auditable"
      ]
    },
```
means that the "modifiedAt" date changed, the list of aspect names changed and three of the properties changed. The `aspectNames` list is the full old list, but the `properties` map is only some of the old custom properties.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/utils/CollectionUtils.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.Collection;
+import java.util.List;
+
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor(access = PRIVATE)
+public final class CollectionUtils
+{
+    public static <T> List<T> difference(Collection<T> first, Collection<T> second)
```
tpage-alfresco, line 41: It would be worth adding some javadoc to clarify that this is `first - second` since it's possible to have the arguments misordered and it's also fairly common to use "difference" when meaning symmetric difference.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/utils/CollectionUtilsTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+class CollectionUtilsTest
+{
+
+    @Test
+    void shouldReturnEmptyListIfCollectionsContainsSameElements()
+    {
+        // given
+        List<String> first = List.of("a", "b", "c");
+        List<String> second = List.of("a", "b", "c");
```
tpage-alfresco, line 43: Since you're working with lists then you could cover more scenarios by testing:
```
Duplication: [a,a]-[a]
Reordering: [a,b,c]-[c,b,a]
```

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/event/UpdateNodeEventMapper.java
```diff
@@ -61,4 +68,50 @@ public UpdateNodeMetadataEvent map(IngestNewNodeEvent event)
...
+            UpdateNodeMetadataEvent updateMetadataEvent,
+            UpdateNodeEvent.FieldDelta<Set<NodeProperty<?>>> customPropertiesDelta)
+    {
+        if (!customPropertiesDelta.updated())
+        {
+            return;
+        }
+
+        List<NodeProperty<?>> toSet = difference(customPropertiesDelta.propertyValue(), customPropertiesDelta.propertyValueBefore());
+        List<NodeProperty<?>> toUnset = difference(customPropertiesDelta.propertyValueBefore(), customPropertiesDelta.propertyValue());
```
tpage-alfresco, line 112: I think the LHS could really be `Set`s as there's no ordering/duplication.

tpage-alfresco (commented):
The only significant issue is that we also need to integrate with the `TransformRequester`, but it's probably best to put that into a second PR.

You already mentioned to me that you need to add some more tests, but everything else is good :+1:

krdabrowski (approved):
LGTM

Pawel-608:
will be done in https://github.com/Alfresco/hxi-connector/pull/34

PR #24, opened 2023-12-13
Put else, catch and finally on their own lines.

tpage-alfresco:
Update the formatter to match what we have across most of our projects.


Pawel-608 (approved):
LGTM

krdabrowski (approved):
LGTM

PR #31, opened 2023-12-18
ACS-6376: introduce property delta

Pawel-608:


live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/metadata/IngestMetadataCommand.java
```diff
@@ -26,9 +26,43 @@
...
+        PropertyDelta<Set<String>> aspectNames,
+        PropertyDelta<Boolean> isFile,
+        PropertyDelta<Boolean> isFolder,
+        PropertyDelta<Long> createdAt,
+        Set<CustomPropertyDelta<?>> properties)
+{
+    public IngestMetadataCommand
+    {
+        ensureNotBlank(nodeId, "Node id cannot be blank");
+        ensureNonNull(name, "Node %s name cannot be blank", nodeId);
```
krdabrowski, line 57: It doesn't match the message, probably should be `NotBlank`

krdabrowski: I'm wondering if it's not possible to use JSR's-303 `@NotNull` and `@NotBlank` or Lombok's `@NonNull` with records 🤔

krdabrowski (approved):
Great job! I only have small remarks.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/metadata/IngestMetadataCommand.java
Pawel-608: it might be possible but it is a little bit tricky - dyk how this validation is performed underhood? I know that:
Lombok
- check is generated somewhere (great, you can easily test that)
- but it is not configurable at all (you can throw just NPE or IAE)
jakarta
- also is not very configurable
- check is performed in some magical way? At least I dont know how it works but I was thinking about it and I found that you have to explicitely call validation (like validator.validate(command))

So lombok is not configureable enough and jakarta is too magical for me (but jakarta annotations are great for example configuration beans etc.)

Pawel-608: So I would leave it at least for now and maybe i future we may try to find some way to use jakarta

Pawel-608: messages were incorrect (all these fields are of `PropertyDelta` type) so I fix them to look like: 
`ensureNonNull(name, "Node %s name delta cannot be null", nodeId);`

PR #35, opened 2023-12-19
Update code formatter - keep record arguments on separate lines

Pawel-608:


tpage-alfresco (commented):
This seems slightly less readable to me:
```
public record ClientData(long modificationTimestamp, String nodeRef)
{}
```
changes to
```
public record ClientData(
    long modificationTimestamp,
    String nodeRef
)
{
}
```

Pawel-608:
@tpage-alfresco do you want me to change it to wrap when more than `n` arguments? (more than 2?) or would you prefer to not merge it at all?

tpage-alfresco:
> @tpage-alfresco do you want me to change it to wrap when more than `n` arguments? (more than 2?) or would you prefer to not merge it at all?

Can we do it based on line length instead? Wrapping when the line exceeds e.g. 140 characters would be best IMO. Otherwise how about wrapping for 5 or more arguments?

FWIW I'd prefer the closing bracket and interfaces to stay on the same line too:
```
    ...
    String replyQueue) implements Serializable
```
rather than
```
    ...
    String replyQueue
) implements Serializable
```

Pawel-608:
meh I think that there is no easy way of setting it up (both put everything into new line if more than n elements and put everything into new lines if line length is > x):
image
so I'll just leave it as it is as it allows for more freedom

PR #38, opened 2023-12-20
Split RepoEventMapper to make it more testable

Pawel-608:


krdabrowski (approved):

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/messaging/in/mapper/property/PropertiesMapperTest.java
```diff
@@ -0,0 +1,232 @@
+package org.alfresco.hxi_connector.live_ingester.messaging.in.mapper.property;
```
tpage-alfresco, line 1: Let's add the same copyright headers to test files too. If the build isn't failing then let's update the plugin to scan all code files.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/messaging/in/utils/RepoEventTestUtils.java
```diff
@@ -0,0 +1,65 @@
...
+    {
+        return ZonedDateTime.ofInstant(Instant.ofEpochMilli(timestamp), ZoneId.of("UTC"));
+    }
+
+    public static void setType(RepoEvent<DataAttributes<NodeResource>> event, EventType type)
+    {
+        given(event.getType()).willReturn(type.getType());
+    }
+
+    public static void setNodeResource(RepoEvent<DataAttributes<NodeResource>> event, NodeResource nodeResource)
```
tpage-alfresco, line 38: This method (and `setNodeResourceBefore`) need a bit of knowledge to use correctly. I'd suggest either or both of:
1. Add some javadoc to explain what they do and how to use them (e.g. `event.getData()` must return a mock).
2. Move the methods back into the test class where they are used (i.e. don't claim they're general purpose).

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/messaging/in/mapper/property/PropertiesMapperTest.java
Pawel-608: meh, I am using two different workspaces and I've set copyright up just on one of them :/ build is failing so is ok from this side

Pawel-608: fixed

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/messaging/in/utils/RepoEventTestUtils.java
Pawel-608: I am moving these methods to individual test classes then

tpage-alfresco (approved):
LGTM! :+1:

PR #51, opened 2024-01-08
ACS-6550 Use final product name 'Alfresco Connector for Hyland Experience Insight'.

tpage-alfresco:
I have left the references in packages as "hxi", but changed the customer-facing locations to either the full name or "Hx Insight." In particular this will affect the git repository (which will need renaming to "hxinsight-connector") and the S3 release bucket (which will also be updated to "hxinsight-connector").


Pawel-608 (approved):
LGTM

krdabrowski (approved):
LGTM

PR #52, opened 2024-01-08
ACS-6498: use api instead of sns

Pawel-608:


tpage-alfresco (approved):
This is great, but the small footprint of the change does highlight our need for some E2E tests.

Pawel-608:
yep I was really suprised pipeline passed😂

live-ingester/pom.xml
```diff
@@ -32,16 +36,18 @@
 			<artifactId>camel-activemq-starter</artifactId>
 		</dependency>
 		<dependency>
-			<groupId>org.apache.camel.springboot</groupId>
-			<artifactId>camel-sjms-starter</artifactId>
+			<groupId>org.apache.camel</groupId>
+			<artifactId>camel-http</artifactId>
```
krdabrowski, line 40: You can use here:
`<dependency>
	<groupId>org.apache.camel.springboot</groupId>
	<artifactId>camel-http-starter</artifactId>
</dependency>`
which already comes with `httpclient` inside

Pawel-608: I think it is not camel that needs it, but `SignedStorageClient` and it previously was supplied as some transitive dependency - so it is better to make direct dependency rather that rely on another transitive one

krdabrowski: Ok, `SignedStorageClient` won't need it any more, but I can remove it in my PR. LGTM

krdabrowski (approved):

PR #54, opened 2024-01-09
ACS-6379: [HxI LI] S3 Credentials - PART II

krdabrowski:
- added storage location gathering
- fixed S3 file uploader - used Camel
- added/fixed tests
- disabled `distribution` Maven profile by default
- cleaned-up Maven config (poms)
- fixed DockerTags implementation and added tests

For more see: https://alfresco.atlassian.net/browse/ACS-6379

PR #55, opened 2024-01-10
ACS-6379 Reduce default build time.

tpage-alfresco:
Split from #54.


tpage-alfresco (commented):
@krdabrowski I can't review "my own" PR, but this LGTM :+1: 

I think we might frequently want to generate docker images locally (but are very unlikely to want zip distributions locally) so we might need to tweak this in the future, but is an improvement on what we have now.

krdabrowski (approved):

krdabrowski:
@tpage-alfresco I've put ✅ for you. Thanks for help!

PR #61, opened 2024-01-12
ACS-6560 Support ACS events with null users.

tpage-alfresco:


live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/metadata/IngestMetadataCommandHandlerTest.java
```diff
@@ -132,4 +132,46 @@ void shouldSetNewlyCreatedNodeMetadataProperties()
...
+    void canSupportEventsWithNullUsers()
+    {
+        // given
+        IngestMetadataCommand command = new IngestMetadataCommand(
+                EVENT_TIMESTAMP,
+                NODE_ID,
+                updated(NODE_NAME),
+                updated(NODE_PRIMARY_ASSOC_Q_NAME),
+                updated(NODE_TYPE),
+                updated(null),
```
krdabrowski, line 145: It's hard to tell what this `null` value stands for - maybe introducing some constant like `CREATED_BY_NULL` would be better 🤔

krdabrowski: Or introducing builder pattern in `IngestMetadaCommand` could be even better 🤔

krdabrowski (approved):
I have only small remarks, overall LGTM

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/metadata/IngestMetadataCommandHandlerTest.java
tpage-alfresco: I've just updated the test to make it a little clearer. I don't have an objection if you want to work on a builder pattern replacement though.

Pawel-608 (changes requested):
I think this is not a good way to handle it :/ as we are introducing null values to the system and you should always avoid it. I would propose one of these solutions:
- create `NULL_USER = "null_user"` constant and use it instead of null (1 line change [there](https://github.com/Alfresco/hxinsight-connector/blob/4973e9c5d8a4a623992f40b6799604938dd37c6b/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/messaging/in/mapper/RepoEventMapper.java#L93)) - the only downside I can see is that hxi might handle null values in the other way than normal strings
- create `User` class and create private constructor that can accept nulls and public one that don't accept nulls and create `NULL_USER = new User(null)` constant - this is better solution imo, but it requires a little bit more work (but it also should be pretty fast)

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/metadata/IngestMetadataCommand.java
```diff
@@ -54,8 +54,6 @@ public record IngestMetadataCommand(
         ensureNonNull(name, "Node %s name delta cannot be null", nodeId);
         ensureNonNull(primaryAssocQName, "Node %s qualified name delta cannot be null", nodeId);
         ensureNonNull(nodeType, "Node %s type delta cannot be null", nodeId);
-        ensureNonNull(createdByUserWithId, "Node %s created by user with nodeId delta cannot be null", nodeId);
-        ensureNonNull(modifiedByUserWithId, "Node %s modified by user with nodeId delta cannot be null", nodeId);
```
Pawel-608, line 58: it actually checks if `propertyDelta` is null not its value or not so it should remain unchanged

tpage-alfresco:
I think it makes sense to go with this solution for now. @Pawel-608 I'm not against you creating a follow-up PR if you have a neater solution, but following the planning meetings today I think we need to move a bit faster.

PR #63, opened 2024-01-15
Ensure "PropertyDelta" wrapper is not null

Pawel-608:


tpage-alfresco (approved):
LGTM - thanks @Pawel-608

PR #66, opened 2024-01-15
ACS-6379: Fixing main POM formatting to meet other POMs style

krdabrowski:
Fixing main POM formatting to meet other POMs style


Pawel-608 (approved):
LGTM

PR #71, opened 2024-01-17
ACS-6379: Fix docker-compose

krdabrowski:
- upgraded version to 3.8
- added healthcheck for alfresco container
- ordered container run: alfresco -> reindexer -> live-indexer

### DIFF
--- a/distribution/src/main/resources/docker-compose/docker-compose.yml
+++ b/distribution/src/main/resources/docker-compose/docker-compose.yml
@@ -1,9 +1,12 @@
-version: "2"
+version: "3.8"
 
 services:
   alfresco:
     image: quay.io/alfresco/alfresco-content-repository:${ALFRESCO_TAG}
-    mem_limit: 1700m
+    deploy:
+      resources:
+        limits:
+          memory: 1700m
     environment:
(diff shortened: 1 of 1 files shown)


Pawel-608 (approved):
LGTM, it'll be really useful, thanks!

distribution/src/main/resources/docker-compose/docker-compose.yml
```diff
@@ -1,9 +1,12 @@
-version: "2"
+version: "3.8"
```
mpichura, line 1: I am not sure if that is compatible with GHA env setup - but assuming that this docker-compose is used in pipeline and tests are green.

mpichura (approved):
LGTM, have look at my comment.

distribution/src/main/resources/docker-compose/docker-compose.yml
krdabrowski: This d-c is not used in pipeline - it's used only for local run of the app

tpage-alfresco (approved):
LGTM. Nice idea to include the reindexing component :+1:

PR #75, opened 2024-01-17
WIP: ACS-6473 Use local app for development by default.

tpage-alfresco:


tpage-alfresco:
For discussion when @Pawel-608 is back. Can we improve our development process to reduce the development cycle by running our applications locally.

This needs:
* Improving to work with multiple local applications (live ingestion, bulk ingestion, prediction handler)
* Documenting so that developers know how they're expected to run the application (we currently have something in the main README)
* Coordination with Adam to see how he'd prefer to run the application for testing purposes.

krdabrowski:
Possible solution could be adding second docker-compose environment adjusted to run Live Ingester out of docker container: https://github.com/Alfresco/hxinsight-connector/pull/76 - of course this is only a proposition and should be extended / documented if we will decide, that this is the way to go

tpage-alfresco:
Closing in favour of #76 

@Pawel-608 We can discuss if you have suggestions when you're back.

PR #76, opened 2024-01-17
ACS-6379: Adding new docker-compose env without ingester service

krdabrowski:
Adding new docker-compose env without ingester service


tpage-alfresco (approved):
This looks good to me. I was thinking of whether we could do something similar without the `extends` by requiring developers to specify one or two yml files like:
```
docker compose -f services.yml up
docker compose -f services.yml -f application.yml up
```

PR #78, opened 2024-01-18
ACS-6473 Fix d-c so that we can download from SFS.

tpage-alfresco:
Also add debug logging for our code to d-c env.


krdabrowski (approved):
LGTM

PR #86, opened 2024-01-22
ACS-6382 Log message when delete event received.

tpage-alfresco:


Pawel-608 (approved):
LGTM

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/delete/DeleteNodeCommand.java
```diff
@@ -0,0 +1,29 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.hxi_connector.live_ingester.domain.usecase.delete;
+
+public record DeleteNodeCommand(long time, String nodeId)
+{}
```
Pawel-608, line 29: I would just add here similar check: 
image

PR #89, opened 2024-01-22
ACS-6380: [HxI LI] added retry logic to PreSignedUrlRequester

krdabrowski:
ACS-6380: [HxI LI] Upload pdf to S3 - PART II - added retry logic to PreSignedUrlRequester

https://alfresco.atlassian.net/browse/ACS-6380

### DIFF
--- a/alfresco-formatter.xml
+++ b/alfresco-formatter.xml
@@ -144,7 +144,7 @@
         <setting id="org.eclipse.jdt.core.formatter.alignment_for_type_arguments" value="0"/>
         <setting id="org.eclipse.jdt.core.formatter.insert_space_after_unary_operator" value="do not insert"/>
         <setting id="org.eclipse.jdt.core.formatter.insert_space_before_opening_paren_in_enum_constant" value="do not insert"/>
-        <setting id="org.eclipse.jdt.core.formatter.alignment_for_arguments_in_annotation" value="0"/>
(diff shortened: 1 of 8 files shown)


Pawel-608 (commented):
Hmmm, I would say we should not use `spring-retry` (btw I did not know that something like this exists so good to know ;)) because there is bult-in camel functionality that actually works the same: 
image
(not sure if it will work but I'm pretty sure that we can configure it to work exactly like we want).

Also mixing spring magic with camel magic may give some bad results (have you tried if it will actually retry delivery?)

krdabrowski:
Using Camel built-in retry (redelivery) logic also came to my mind - earlier it didn't work for me, but now when I'm testing this seems to work fine, so I had to do sth wrong.
Regarding Spring Retry I've tested it (even wrote integration tests), everything works fine and it's very easy to test - in the opposite to Camel logic. At least I didn't found yet an easy way how to test these redeliveries, but I'm still playing with it.

tpage-alfresco:
On the subject of Camel retry vs Spring retry - FWIW we're using a combination of both in the ESC. I think this is slightly easier to understand than some of the Camel exception handling in ESC, but maybe I'm not comparing like-for-like. Anyway - I'm happy with this the way it is, if you want to rework it with Camel then I'm happy to see that too.

Pawel-608 (approved):
@tpage-alfresco @krdabrowski I am approving it then

PR #95, opened 2024-01-25
ACS-6383 E2E test for delete request.

tpage-alfresco:


azakrzewski-hy (approved):
LGTM

Pawel-608 (approved):

live-ingester/pom.xml
```diff
@@ -53,6 +53,11 @@
             <artifactId>camel-mock-starter</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-test-spring-junit5</artifactId>
```
krdabrowski, line 58: Is this dependency used anywhere?

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/util/E2ETestBase.java
```diff
@@ -0,0 +1,106 @@
...
+import org.testcontainers.utility.DockerImageName;
+import org.wiremock.integrations.testcontainers.WireMockContainer;
+
+@SpringBootTest(properties = {
+        "logging.level.org.alfresco=DEBUG"
+})
+@Testcontainers
+@DirtiesContext // Forces framework to kill application after tests (i.e. before testcontainers die).
+@NoArgsConstructor(access = PROTECTED)
+public class E2ETestBase
```
krdabrowski, line 58: This probably could be `abstract` as it's not intended to contain/run any test cases, WDYT?

pom.xml
```diff
@@ -137,6 +137,18 @@
...
             </dependency>
+            <dependency>
+                <groupId>org.apache.camel.springboot</groupId>
+                <artifactId>camel-mock-starter</artifactId>
+                <version>${spring-camel.version}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.camel</groupId>
+                <artifactId>camel-test-spring-junit5</artifactId>
```
krdabrowski, line 148: This is not necessary - both explicit declarations may be skipped here, as both are covered by `camel-spring-boot-dependencies` bom.

krdabrowski (commented):
I was catching-up and have found few things, which are unclear to me. In general great job!

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/util/E2ETestBase.java
tpage-alfresco: I initially had it as abstract, but PMD pointed out that it was neater to have a protected constructor to prevent accidental instantiation. I don't think it really matters though as no one is going to try to create one of these.

krdabrowski: Oh, got it... I'm fine with this approach

PR #96, opened 2024-01-25
ACS-6671 Fix properties format in Hx Insight request.

tpage-alfresco:
Also add E2E test for this. Currently the E2E test does not check anything about the ATS request, but we should also include that in the future.


Pawel-608 (approved):
LGTM

PR #100, opened 2024-01-26
ACS-6673: Bulk Ingester - log every node id

Pawel-608:


bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/BulkIngesterConfig.java
```diff
@@ -23,18 +23,23 @@
...
-@SuppressWarnings("PMD.UseUtilityClass")
-public class BulkIngesterApplication
+import jakarta.validation.constraints.Positive;
+import jakarta.validation.constraints.PositiveOrZero;
+
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+@Data
```
krdabrowski, line 36: WDYT about getting rid of `get` prefix and making the accessors fluent? Also do we need setters here?

krdabrowski (approved):
Overall LGTM, I only have one remark

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/BulkIngesterConfig.java
Pawel-608: hmm, making getters fluent: +, but I think we have setters because it cannot fill properties without them. I saw in your PR you used records - so I maybe will try to also use records? imo it was really nice idea

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/BulkIngesterConfig.java
```diff
@@ -23,18 +23,23 @@
...
+public class BulkIngesterConfig
 {
+    @PositiveOrZero
+    private long fromId;
 
-    public static void main(String[] args)
-    {
-        SpringApplication.run(BulkIngesterApplication.class, args);
-    }
+    @Positive private long toId;
```
tpage-alfresco, line 44: I'm surprised the formatter doesn't have a preference over inline annotations or multiline. FWIW I prefer annotations on their own line.

Pawel-608: I think formatter did it :/ I'll try to move it to separate line but idk if it'll work

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/BulkIngesterApplication.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.context.properties.ConfigurationPropertiesScan;
+
+import org.alfresco.elasticsearch.db.connector.config.DatabaseConfig;
+import org.alfresco.hxi_connector.bulk_ingester.processor.BulkIngestionProcessor;
+
+@RequiredArgsConstructor
+@SpringBootApplication
+@ImportAutoConfiguration(DatabaseConfig.class)
+@ConfigurationPropertiesScan("org.alfresco.hxi_connector.bulk_ingester")
```
krdabrowski, line 42: You can remove annotation value if it matches current class package

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/BulkIngesterConfig.java
krdabrowski: Spring can populate the values by constructor using so called `@ConstructorBinding` - you don't have to even annotate the class with that, just make the fields `final`

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/BulkIngesterConfig.java
tpage-alfresco: That would be a plot twist if the formatter likes a mixture of same line/separate line :yum:

tpage-alfresco:
Overall this looks great. It gives us a big step in the right direction, and we can add more features step by step :+1:

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/BulkIngesterConfig.java
Pawel-608: I'll read about it - for now I used record and imo it looks good ;)

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/BulkIngesterConfig.java
Pawel-608: yep it is formatter thing :/

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/BulkIngesterApplication.java
Pawel-608: fixed!

tpage-alfresco (approved):
LGTM. We saw the build should pass after this point :+1:

PR #103, opened 2024-01-29
ACS-6676 Handle update from null to null correctly.

tpage-alfresco:


tpage-alfresco:
@Pawel-608 While working on this I found `CategoryPropertyResolver` which seems to be unused. Can we get rid of this class now or is it meant to be wired in somewhere?

PR #105, opened 2024-01-29
Ensure categories and tags mapping is applied

Pawel-608:


tpage-alfresco (changes requested):
Ah - sorry, would it be possible to make this change in the Create test instead? This property is already being used to test the weird `null` to `null` case.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/UpdateRequestIntegrationTest.java
```diff
@@ -124,4 +124,244 @@ void testUpdateRequest()
...
+
+        // when
+        containerSupport.raiseRepoEvent(repoEvent);
+
+        // then
+        String expectedBody = """
+                {
+                  "objectId" : "d71dd823-82c7-477c-8490-04cb0e826e65",
+                  "eventType" : "update"
+                }""";
```
tpage-alfresco, line 229: Hmm - this is a weird edge case. I guess really we shouldn't send anything in this case, but actually it should never happen since I think the modification date will always be updated.

Pawel-608: I can actually add some random property like title updated wdyt?

Pawel-608: so we'll have 
image

tpage-alfresco (approved):
Nice collection of test cases :+1:

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/UpdateRequestIntegrationTest.java
tpage-alfresco: I'm happy either way. Like this we at least check what happens when there's no real change, even if it's a bit weird to tell Hx Insight "there was no change to this node".

PR #108, opened 2024-01-30
Remove unused dependencies.

tpage-alfresco:
These were added while trying to get the e2es working, but are not needed: https://github.com/Alfresco/hxinsight-connector/pull/95/files#r1469657862


krdabrowski (approved):
LGTM

PR #109, opened 2024-01-30
ACS-6682: Map nodes we get from database to something we can send in the event

Pawel-608:


bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoNodeMapper.java
```diff
@@ -0,0 +1,108 @@
...
+
+@Slf4j
+@RequiredArgsConstructor(access = PRIVATE)
+public class AlfrescoNodeMapper
+{
+    public static final String NAME_PROPERTY = "name";
+    public static final String CONTENT_PROPERTY = "content";
+    private static final Set<String> PREDEFINED_PROPERTIES = Set.of(NAME_PROPERTY, CONTENT_PROPERTY);
+
+    private final AlfrescoNode alfrescoNode;
```
tpage-alfresco, line 55: I think this would be easier to follow if we just pass `AlfrescoNode` rather than store it as a field. We would be able to remove one of `map` or `performMapping`.

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/model/Node.java
```diff
@@ -0,0 +1,46 @@
...
+package org.alfresco.hxi_connector.bulk_ingester.processor.model;
+
+import java.io.Serializable;
+import java.time.ZonedDateTime;
+import java.util.Map;
+import java.util.Set;
+
+import org.alfresco.elasticsearch.db.connector.model.ContentMetadata;
+
+public record Node(
```
tpage-alfresco, line 36: Could we instead map the DB data to an `IngestMetadataCommand` and an `IngestContentCommand` since these are already part of our internal model? Do you think that would be too much for one step?

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoNodeMapper.java
```diff
@@ -0,0 +1,108 @@
...
+    {
+        String nodeId = alfrescoNode.getNodeRef();
+        String type = alfrescoNode.getType().getLocalName();
+        String creatorId = alfrescoNode.getCreator();
+        String modifierId = alfrescoNode.getModifier();
+        Set<String> aspectNames = alfrescoNode.getAspects().stream().map(QName::getLocalName).collect(Collectors.toSet());
+        ZonedDateTime createdAt = alfrescoNode.getCreatedAt();
+        Map<String, Serializable> allProperties = calculateAllProperties();
+
+        String name = (String) allProperties.get(NAME_PROPERTY);
```
tpage-alfresco, line 72: It looks easier to just leave the name property in with the other properties. I don't think we use it for anything.

Pawel-608: So should we put it in custom properties then (in IngestMetadataCommand etc.?)

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoNodeMapper.java
Pawel-608: done

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/model/Node.java
Pawel-608: I dont think it makes any sense as we have to send the events anyway so it will be easier to have some custom class we can easily serialise/deserialise

tpage-alfresco: I was thinking we could serialise/deserialise those command classes. I'm happy with this approach too, but eventually we know we want to get those two classes so I thought it might be simpler overall.

Pawel-608: there is this `PropertyDelta` wrapper around all properties in `IngestMetadataCommand` so imo it will be hard to write some serialization logic that makes sense :/ Also we would have to somehow import `live-ingester` to `bulk-ingester` which may cause with some coupling we would not like to have So I think leaving this class is just much simpler approach

Pawel-608: we can always refactor it later if needed

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapper.java
```diff
@@ -0,0 +1,201 @@
...
+            String storeIdentifier = storeSplit[0];
+            String id = storeSplit[1];
+
+            return of((Serializable) Map.of(
+                    "id", id,
+                    "storeRef", Map.of("identifier", storeIdentifier, "protocol", protocol)));
+        }
+        catch (Exception e)
+        {
+            log.error(format("Cannot deserialize noderef property %s with value %s for node %s", propertyName, persistedValue, alfrescoNode.getId()), e);
```
tpage-alfresco, line 124: Slf4j already has token handling:
```suggestion
            log.error("Cannot deserialize noderef property {} with value {} for node {}", propertyName, persistedValue, alfrescoNode.getId(), e);
```

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
```diff
@@ -0,0 +1,489 @@
...
+                Set.of(
+                        createNodeProperty(createdAt, dateValue(createdAtString))));
+
+        // when
+        var property = new AlfrescoPropertyMapper(alfrescoNode, createdAt).performMapping();
+
+        // then
+        assertTrue(property.isPresent());
+        assertEquals(createdAt, property.get().getKey());
+        assertEquals(createdAtValue, property.get().getValue());
```
Pawel-608, line 327: and also I am wondering if we actually need it

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapper.java
```diff
@@ -0,0 +1,201 @@
...
+            {
+            case CONTENT_DATA_ID -> getContentValue(persistedValue);
+            case DATE -> getDateValue(persistedValue);
+            case NODEREF -> getNodeRefValue(persistedValue);
+            default -> of(persistedValue);
+            });
+        }
+        catch (Exception e)
+        {
+            log.warn(format("Error occurred while trying to map property %s value %s of node %s", propertyName, propertyValue, alfrescoNode.getId()), e);
```
tpage-alfresco, line 99:
```suggestion
            log.warn("Error occurred while trying to map property {} value {} of node {}", propertyName, propertyValue, alfrescoNode.getId(), e);
```

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapper.java
```diff
@@ -0,0 +1,201 @@
...
+    {
+        try
+        {
+            ObjectInput in = new ObjectInputStream(new ByteArrayInputStream(propertyValue.getSerializableValue()));
+
+            return of((Serializable) in.readObject());
+        }
+        catch (ClassNotFoundException | IOException e)
+        {
+            log.error(format("Cannot deserialize property %s value %s. Node: %s", propertyName, propertyValue, alfrescoNode.getId()), e);
```
tpage-alfresco, line 196:
```suggestion
            log.error("Cannot deserialize property {} value {}. Node: {}", propertyName, propertyValue, alfrescoNode.getId(), e);
```

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/model/Node.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package org.alfresco.hxi_connector.bulk_ingester.processor.model;
+
+import java.io.Serializable;
+import java.time.ZonedDateTime;
+import java.util.Map;
+import java.util.Set;
+
+import org.alfresco.elasticsearch.db.connector.model.ContentMetadata;
```
tpage-alfresco, line 34: We could map this to some internal object too - WDYT?

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
```diff
@@ -0,0 +1,489 @@
...
+@SuppressWarnings({"PMD.TooManyMethods", "PMD.LinguisticNaming"})
+class AlfrescoPropertyMapperTest
+{
+    @Test
+    void shouldMapJustPropertiesWithGivenName()
+    {
+        // given
+        AlfrescoNode alfrescoNode = new AlfrescoNode();
+
+        String description = "description";
```
tpage-alfresco, line 70: I think this would be slightly clearer if the variable was called `propertyName`.

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
```diff
@@ -0,0 +1,489 @@
...
+                        createNodeProperty("title"),
+                        createNodeProperty(description, stringValue(descriptionText))));
+
+        // when
+        var property = new AlfrescoPropertyMapper(alfrescoNode, description).performMapping();
+
+        // then
+        assertTrue(property.isPresent());
+        assertEquals(description, property.get().getKey());
+        assertEquals(descriptionText, property.get().getValue());
```
tpage-alfresco, line 84: Can we replace the three statements with a single equality check like:
```
assertEquals(expectedOptional, property);
```

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
```diff
@@ -0,0 +1,489 @@
...
+
+    @Test
+    void shouldProcessPropertiesWithSameNameTogether()
+    {
+        // given
+        AlfrescoNode alfrescoNode = new AlfrescoNode();
+
+        String description = "description";
+        String usDescription = "French fries recipe";
+        String ukDescription = "Chips recipe";
```
tpage-alfresco, line 95: Nice example! :fries:

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
```diff
@@ -0,0 +1,489 @@
...
+                        createNodeProperty(description, "en_US_", stringValue(usDescription)),
+                        createNodeProperty(description, "en_UK_", stringValue(ukDescription))));
+
+        // when
+        var property = new AlfrescoPropertyMapper(alfrescoNode, description).performMapping();
+
+        // then
+        assertTrue(property.isPresent());
+        assertEquals(description, property.get().getKey());
+        assertThat(List.of(usDescription, ukDescription)).hasSameElementsAs((Iterable<String>) property.get().getValue());
```
tpage-alfresco, line 109: I think the two arguments are backwards - `assertThat(actual).hasSameElementsAs(actual)`

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
```diff
@@ -0,0 +1,489 @@
...
+    }
+
+    @Test
+    void shouldProcessLongProperties()
+    {
+        // given
+        AlfrescoNode alfrescoNode = new AlfrescoNode();
+
+        String legsCount = "legsCount";
+        long legsCountValue = 3;
```
tpage-alfresco, line 181: Using long for a count of the number of legs :scream:

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
```diff
@@ -0,0 +1,489 @@
...
+                Set.of(
+                        createNodeProperty(legLength, floatValue(legLengthValue))));
+
+        // when
+        var property = new AlfrescoPropertyMapper(alfrescoNode, legLength).performMapping();
+
+        // then
+        assertTrue(property.isPresent());
+        assertEquals(legLength, property.get().getKey());
+        assertEquals(legLengthValue, property.get().getValue());
```
tpage-alfresco, line 215: This is fine because we know we're just passing the float values through the tested method (same for the next test with doubles).

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
```diff
@@ -0,0 +1,489 @@
...
+    }
+
+    @Test
+    void shouldProcessContentComplexProperty()
+    {
+        // given
+        AlfrescoNode alfrescoNode = new AlfrescoNode();
+
+        String content = "content";
+        long contendId = 1;
```
tpage-alfresco, line 288: Typo - `contentId`

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
tpage-alfresco: I think we agreed we'd just send timestamps to Hx Insight, but if we can consistently send ISO format instead then great.

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
```diff
@@ -0,0 +1,488 @@
...
+                        createNodeProperty(copyOf, nodeRefValue(copyOfValue))));
+
+        // when
+        var property = new AlfrescoPropertyMapper(alfrescoNode, copyOf).performMapping();
+
+        // then
+        assertTrue(property.isPresent());
+        assertEquals(copyOf, property.get().getKey());
+        assertEquals(copyOfExpectedValue, property.get().getValue());
+    }
```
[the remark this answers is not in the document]
tpage-alfresco: I expect that Hx Insight will not use the store ref at all. They might possibly use the UUID to match up with other nodes although we're going to send associations in a different format. Overall I don't think it matters too much, so let's just do something for now and change it later if there's a request for it.

tpage-alfresco (approved):
I added quite a few comments, but nothing major. Looks like a really good start to the bulk ingester :+1:

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
krdabrowski: I think you meant `assertThat(actual).hasSameElementsAs(expected)`

krdabrowski (approved):
Nice job!

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
Pawel-608: in case someone would like to put here number of legs of `centipede` from "James and the Giant Peach" ;)

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
Pawel-608: good point, I'll convert it to timestamp then

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
Pawel-608: fixed!

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/model/Node.java
Pawel-608: hmm, you're right I'll do it

Pawel-608: done!

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapper.java
Pawel-608: hmmm, I just checked it and it actually works - thanks for suggestion!

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapper.java
Pawel-608: fixed!

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapper.java
Pawel-608: fixed!

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
Pawel-608: that's right, fixed

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapper.java
Pawel-608: fixed

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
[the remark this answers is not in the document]
Pawel-608: done

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
Pawel-608: done

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoPropertyMapperTest.java
Pawel-608: good point! fixed

PR #111, opened 2024-01-31
ACS-6380: reorganize configuration properties

krdabrowski:
https://alfresco.atlassian.net/browse/ACS-6380


live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/IntegrationConfig.java
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.validation.annotation.Validated;
+
+import org.alfresco.hxi_connector.live_ingester.adapters.config.properties.Ingester;
+import org.alfresco.hxi_connector.live_ingester.adapters.config.properties.Repository;
+import org.alfresco.hxi_connector.live_ingester.adapters.config.properties.Storage;
+import org.alfresco.hxi_connector.live_ingester.adapters.config.properties.Transform;
+
+@Configuration
+@EnableConfigurationProperties
+public class IntegrationConfig
```
tpage-alfresco, line 45: Do we need this class? It seems like we're using the `Properties` inner class everywhere, which suggests that would be more useful as a class of it's own using the path `alfresco.integration.properties`.

Pawel-608 (approved):
LGTM, great work! it is way more readable now

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/IntegrationConfig.java
krdabrowski: I had issues with this approach earlier due two beans, which Spring created:
`alfresco.integration-org.alfresco.hxi_connector.live_ingester.adapters.config.IntegrationProperties`
`integrationProperties`
but right now it seems to work fine...
Possibly I had some misconfiguration related with `@EnableConfigurationProperties` or `@ConfigurationPropertiesScan` 🤔 

As it works fine I've fixed that according to your suggestion.

live-ingester/src/main/resources/application.yml
```diff
@@ -14,28 +14,26 @@ hyland:
...
+      location:
+        endpoint: "${hyland.experience.insight.api.url.storage-location-request}\
+          ?httpMethod=POST\
+          &authMethod=Basic\
+          &authUsername=${hyland.experience.insight.api.username}\
+          &authPassword=${hyland.experience.insight.api.password}\
+          &authenticationPreemptive=true\
+          &throwExceptionOnFailure=false"
+    ingester:
+      endpoint: ${hyland.experience.insight.api.url.ingest}?httpMethod=POST
```
tpage-alfresco, line 39: I think this should be under `hyland.experience` rather than `alfresco`.

live-ingester/src/main/resources/application.yml
```diff
@@ -14,28 +14,26 @@ hyland:
...
-      endpoint: activemq:queue:acs-repo-transform-request?jmsMessageType=Text
-      timeout: 20000
-    response:
-      queueName: org.alfresco.hxinsight-connector.transform.response
-      endpoint: activemq:queue:${alfresco.transform.response.queueName}
-    shared-file-store:
-      host: http://localhost
-      port: 8099
+      location:
+        endpoint: "${hyland.experience.insight.api.url.storage-location-request}\
```
tpage-alfresco, line 31: I think this should be under `hyland.experience` rather than `alfresco`.

live-ingester/src/main/resources/application.yml
```diff
@@ -14,28 +14,26 @@ hyland:
 
 alfresco:
   integration:
+    repository:
+      endpoint: activemq:topic:alfresco.repo.event2
```
tpage-alfresco, line 18: This change makes it clearer what the property is for :+1:

tpage-alfresco (commented):
I think this contains two slightly different things:
* Reorganisation of the existing properties
* Addition of code for handling retry properties

From what I can see then the retry properties weren't added in this PR.

Consequently I suggest splitting it into two PRs with the refactor in this one and the retry properties in the next.

live-ingester/src/main/resources/application.yml
```diff
@@ -14,28 +14,26 @@ hyland:
...
+    transform:
+      request:
+        endpoint: activemq:queue:acs-repo-transform-request?jmsMessageType=Text
+        timeout: 20000
+      response:
+        endpoint: activemq:queue:${alfresco.transform.response.queueName}
+        queueName: org.alfresco.hxinsight-connector.transform.response
```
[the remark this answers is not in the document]
krdabrowski: Fixed

live-ingester/src/main/resources/application.yml
krdabrowski: Fixed

live-ingester/src/main/resources/application.yml
krdabrowski: Definitely I agree

krdabrowski:
@tpage-alfresco I've refactored the PR - it's ready for review again

tpage-alfresco (approved):
Looks good Krys! Thanks for splitting this out :+1:

live-ingester/src/main/resources/application.yml
```diff
@@ -2,40 +2,36 @@ camel:
...
     response:
-      queueName: org.alfresco.hxinsight-connector.transform.response
-      endpoint: activemq:queue:${alfresco.transform.response.queueName}
+      queue-name: org.alfresco.hxinsight-connector.transform.response
+      endpoint: activemq:queue:${alfresco.transform.response.queue-name}
     shared-file-store:
       host: http://localhost
       port: 8099
+hyland-experience:
+  insight:
```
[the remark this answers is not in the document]
krdabrowski: Ok, I will leave it as it is. In future we can fix it if required

PR #127, opened 2024-02-05
Add BulkIngesterNodeRepositoryTest

Pawel-608:


bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/repository/BulkIngesterNodeRepositoryTest.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Test
+    void ensureNodeStreamIsLazyEvaluated()
+    {
+        // given
+        ListAppender<ILoggingEvent> testAppender = createTestLogsAppender();
+
+        addAppenderForLogger(testAppender, BulkIngesterNodeRepository.class);
+        addAppenderForLogger(testAppender, BulkIngesterNodeRepositoryTest.class);
```
krdabrowski, line 93: Interesting approach with this verifying against logger, but are this two declarations important from test case point of view? Maybe it would be better to hide them within `createTestLogsAppender()` and make the test a bit simpler?

krdabrowski (approved):
LGTM

PR #128, opened 2024-02-05
ACS-6718: Put name property in custom properties

Pawel-608:


live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/in/mapper/property/PropertiesMapper.java
```diff
@@ -77,6 +73,49 @@ public Set<CustomPropertyDelta<?>> calculateCustomPropertiesDelta(RepoEvent<Data
...
+    {
+        if (shouldNotUpdateField(event, NodeResource::getName))
+        {
+            return Stream.empty();
+        }
+
+        return ofNullable(event.getData().getResource().getName())
+                .stream()
+                .filter(Objects::nonNull)
+                .map(name -> CustomPropertyDelta.updated(NAME_PROPERTY_KEY, name));
```
Pawel-608, line 116: should we use "fully qualified name" or simple name or maybe both? wdyt? I would say we should use just fully qualified one and ignore simple name

krdabrowski: How does it looks in case of other properties? Do we use the "fully qualified name"?

Pawel-608: so for other "custom" properties keys we use fully qualified names (as it comes in event), but idk if any other property can have two different "values", I'll maybe show an example:

we would have
`cm:title` not `title` property key and its value can be `some interesting title`
for name we would have
`cm:name` not `name` as property key but as value we could have both `cm:my_file_name` and `my_flie_name`, and the question is, whether we want/need to store both or whether one is enough (and in that case which one we prefer)

tpage-alfresco: I would stick with just `cm:name` as the only name. I asked a very similar question to the Hx Insight team and they weren't interested in having any "well known" field names other than an id.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/metadata/model/PredefinedNodeMetadataPropertyTest.java
```diff
@@ -39,13 +39,13 @@ class PredefinedNodeMetadataPropertyTest
...
-        String name = "test name";
+        String name = "test user";
 
         // when
-        NodeProperty<String> nodeProperty = PredefinedNodeMetadataProperty.NAME.withValue(name);
+        NodeProperty<String> nodeProperty = PredefinedNodeMetadataProperty.CREATED_BY_USER_WITH_ID.withValue(name);
 
         // then
-        assertEquals(PredefinedNodeMetadataProperty.NAME.getName(), nodeProperty.name());
+        assertEquals(PredefinedNodeMetadataProperty.CREATED_BY_USER_WITH_ID.getName(), nodeProperty.name());
```
tpage-alfresco, line 48: Do we want to treat created by differently? In the DB it's just `cm:creator`

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/in/mapper/property/PropertiesMapperTest.java
```diff
@@ -144,12 +147,64 @@ void shouldHandleAllCustomPropertiesUpdated_NodeCreated()
...
+        // given
+        String name = "some name";
+
+        RepoEvent<DataAttributes<NodeResource>> event = mock();
+
+        setType(event, NODE_CREATED);
+
+        NodeResource nodeResource = NodeResource.builder()
+                .setName(name)
+                .setProperties(Map.of())
```
tpage-alfresco, line 169: Do we need this line?

tpage-alfresco (approved):

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/in/mapper/property/PropertiesMapper.java
krdabrowski: The most safe approach would be to preserve both values as I'm note sure if HxI will require them both or one of them 🤔 Maybe @tpage-alfresco can tell something more in this topic

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/in/mapper/property/PropertiesMapperTest.java
Pawel-608: hmm, it throws npe if properties are not set but maybe actually I should fix the code instead of setting it to empty map here🤔

Pawel-608: fixed

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/in/mapper/property/PropertiesMapper.java
Pawel-608: what about values? `cm:my_file_name`/ `my_file_name`/ both?

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/metadata/model/PredefinedNodeMetadataPropertyTest.java
Pawel-608: I think we have custom column for that🤔

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/in/mapper/property/PropertiesMapper.java
tpage-alfresco: Pawel clarified that `cm:my_file_name` is coming from `primaryAssocQName`. This field doesn't give any additional useful information that isn't already present in other fields. I would stick with just `cm:name` and so the value should just be `my_file_name` without a prefix.

PR #129, opened 2024-02-05
ACS-6719: Handle bulk-ingester events [metadata]

Pawel-608:


live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/bulk_ingester/BulkIngesterEventProcessor.java
```diff
@@ -0,0 +1,76 @@
...
+@Component
+@RequiredArgsConstructor
+public class BulkIngesterEventProcessor
+{
+    private final IngestMetadataCommandHandler ingestMetadataCommandHandler;
+
+    public void process(BulkIngesterEvent event)
+    {
+        IngestMetadataCommand ingestMetadataCommand = new IngestMetadataCommand(
+                event.createdAt(),
```
Pawel-608, line 54: This filed is "event time" field - Should we use node creation timestamp or event timestamp?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/bulk_ingester/BulkIngesterEventProcessor.java
```diff
@@ -0,0 +1,76 @@
...
+                event.nodeId(),
+                false,
+                PropertyDelta.updated(event.type()),
+                PropertyDelta.updated(event.creatorId()),
+                PropertyDelta.updated(event.modifierId()),
+                PropertyDelta.updated(event.aspectNames()),
+                PropertyDelta.updated(false),
+                PropertyDelta.updated(false),
```
Pawel-608, line 62: these two are "isFile" and "isFolder" properties - we don't have them now - I would add them later

Pawel-608: it probably will not be too easy🤔 They are calculated in repo basing on node type (but it is not `type == "folder"` check rather `is type subtype of "folder"`)

krdabrowski: I'm ok with solving this in separate PR, did you raised a ticket for that?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/bulk_ingester/BulkIngesterRouteBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.camel.builder.RouteBuilder;
+import org.springframework.stereotype.Component;
+
+import org.alfresco.hxi_connector.live_ingester.adapters.config.IntegrationProperties;
+import org.alfresco.hxi_connector.live_ingester.adapters.messaging.bulk_ingester.model.BulkIngesterEvent;
+import org.alfresco.hxi_connector.live_ingester.domain.exception.LiveIngesterRuntimeException;
+
+@Component
+@RequiredArgsConstructor
+public class BulkIngesterRouteBuilder extends RouteBuilder
```
krdabrowski, line 44: This suffix: `RouteBuilder` doesn't tell anything IMO, I would rather see something more self-explanatory, maybe `BulkIngestListener` or `BulkIngesterEventListener`?

krdabrowski (approved):
LGTM, I have only one remark

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/bulk_ingester/BulkIngesterEventProcessor.java
Pawel-608: hmm, I just discovered we don't use this property at all - I'll remove it then (but maybe in next pr)

Pawel-608: https://alfresco.atlassian.net/browse/ACS-6727

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/bulk_ingester/BulkIngesterEventProcessor.java
Pawel-608: I didn't yet - maybe we don't have to do it anyway? @tpage-alfresco do you think that these two properties are necessary? because we would need info about types structure to calculate them (https://github.com/Alfresco/alfresco-community-repo/blob/71b69529c660ccc9eeecd22791999ba0b02b00f7/amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/report/generator/transfer/TransferReportGenerator.java#L117) - but maybe we will have required information anyway because we'll need it to calculate qnames of properties?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/bulk_ingester/BulkIngesterRouteBuilder.java
Pawel-608: Good point! fixed

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/bulk_ingester/BulkIngesterEventProcessor.java
tpage-alfresco: I don't think they're necessary. Hx Insight can make predictions based on the type and I doubt knowing `isFile`/`isFolder` will make a significant difference.

Pawel-608: I'll remove them in this case

Pawel-608: https://alfresco.atlassian.net/browse/ACS-6734

PR #131, opened 2024-02-06
ACS-6719: Handle bulk ingester events [content]

Pawel-608:


tpage-alfresco (approved):
LGTM! We need tests for the bulk ingester code.

PR #133, opened 2024-02-07
ACS-6727: Add namespace prefixes to properties [bulk-ingester]

Pawel-608:


bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/config/PredefinedNamespacePrefixMapper.java
```diff
@@ -0,0 +1,81 @@
...
+@Data
+@Slf4j
+@Validated
+@Configuration
+@ConfigurationProperties
+@PropertySource(value = "${alfresco.bulk.ingest.namespace-prefixes-mapping}", factory = PredefinedNamespacePrefixMapper.JsonPropertySourceFactory.class)
+public class PredefinedNamespacePrefixMapper implements NamespacePrefixMapper
+{
+    @NotNull private Map<String, String> prefixUriMap;
+
```
Pawel-608, line 55: I copied most of this class from elasticsearch-connector but now I am wondering if maybe it wouldn't be better to just manually read this file containing prefixes mappings and mark this class just as @Configuration or something because current implementation is misleading for intellij: 
image
wdyt?

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/TestNamespaceToPrefixMapper.java
```diff
@@ -0,0 +1,47 @@
...
+ * #L%
+ */
+
+package org.alfresco.hxi_connector.bulk_ingester.processor.mapper;
+
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+@SuppressWarnings({"PMD.TestClassWithoutTestCases"})
+class TestNamespaceToPrefixMapper implements NamespacePrefixMapper
```
tpage-alfresco, line 33: Is this class necessary? Can't we just mock the mapping in the tests?

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/config/PredefinedNamespacePrefixMapper.java
tpage-alfresco: How about proposing something as a separate PR? This PR gets the functionality working, so is valuable by itself.

bulk-ingester/src/main/resources/application.yml
```diff
@@ -15,3 +15,4 @@ alfresco.bulk.ingest:
   repository:
     pageSize: 100
   endpoint: activemq:queue:bulk-ingester-events
+  namespace-prefixes-mapping: classpath:namespace-prefixes.json
```
tpage-alfresco, line 18: We should update the README to document this file and where customers need to put it.

tpage-alfresco (approved):
LGTM. We should also have some documentation but it can go in a separate PR.

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/TestNamespaceToPrefixMapper.java
Pawel-608: it was much easier to create this test implementation, than trying to mock it as I had to implement just one method:
image

PR #138, opened 2024-02-07
Fix invalid ID in file upload request

krdabrowski:


live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/TransformResponse.java
```diff
@@ -23,22 +23,19 @@
...
-@NoArgsConstructor
-@AllArgsConstructor
-public class TransformResponse
-{
-    @NotBlank
-    private String targetReference;
-}
+public record TransformResponse(
+        @NotBlank String targetReference,
+        @NotNull @JsonDeserialize(using = ClientDataDeserializer.class) ClientData clientData)
```
Pawel-608, line 40: I think you can add this deserializer in `MappingConfig` instead

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/model/ClientDataDeserializer.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.hxi_connector.live_ingester.adapters.messaging.transform.model;
+
+import org.alfresco.hxi_connector.live_ingester.adapters.config.jackson.RawJsonDeserializer;
+
+public class ClientDataDeserializer extends RawJsonDeserializer<ClientData>
```
Pawel-608, line 30: I would move this class to config/jacson package tbh

Pawel-608 (approved):
LGTM, just small remarks regarding jackson stuff

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/TransformResponse.java
krdabrowski: Annotations are more descriptive IMO and tell straightforwardly that something is handled differently, in the other hand I don't mind hiding this in `MappingConfig` and simplifying record 🤔

tpage-alfresco (approved):
LGTM

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/TransformResponse.java
krdabrowski: It turned out, that when I register the deserilizer globally then the `RawJsonDeserializer` which uses global `ObjectMapper` calls itself again in some kind of circular relation but without the JSON value and fails. I will leave it as it is and we can change it in the future.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/model/ClientDataDeserializer.java
krdabrowski: Done

PR #141, opened 2024-02-08
ACS-6685 Handle update that removes content.

tpage-alfresco:


Pawel-608 (commented):
First question is - do we have to do it this way😂? I think we are not setting "cm:content" property anywhere right? So why we would like to unset it? Maybe it is better to ask guys from hxi to create some separate endpoint that will allow us to send them information that content was deleted? Also imo it is not the right way of doing it - you insert very specific behaviour to a generic `IngestMetadataCommand` class - because of that it grows in size, you have to change all tests related to this functionality and it starts to be hard to understand. Instead I would propose to create new `DeleteContentMetadataCommand` and handle it separately - you can even write this logic in `IngestMetadataCommandHandler`, but it will be in separate method

tpage-alfresco:
The `cm:content` property is not yet being sent to Hx Insight. It will happen as part of https://alfresco.atlassian.net/browse/ACS-6381.

Update events from the Repo trigger two actions:
* Update HxInsight (with the "metadata")
* Kick off the transformation

image

When the content is deleted from a node there is no need for a transformation and so we can immediately notify Hx Insight that the content has been removed. Since we're already sending the metadata then we should include the content deletion in the same message.

> Maybe it is better to ask guys from hxi to create some separate endpoint that will allow us to send them information that content was deleted?

If you want to discuss the API design then the page is here and still marked as open for feedback:
https://alfresco.atlassian.net/wiki/spaces/HxAI/pages/1511425263/Ingestion+JSON+Payload+Format

> ...you insert very specific behaviour to a generic IngestMetadataCommand class

I think this is the right generic way to handle content deletion (as part of an update message to Hx Insight). Currently we're only supporting a single piece of content, but in the future we plan to support other content properties too - maybe we can wrap the property name `cm:content` so that we don't need to know about it internally. We could also pass the property name in the `clientData` part of the ATS request.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/metadata/IngestMetadataCommand.java
```diff
@@ -42,7 +42,8 @@ public record IngestMetadataCommand(
         PropertyDelta<String> modifiedByUserWithId,
         PropertyDelta<Set<String>> aspectNames,
         PropertyDelta<Long> createdAt,
-        Set<CustomPropertyDelta<?>> properties)
+        Set<CustomPropertyDelta<?>> properties,
+        boolean contentRemoved)
```
tpage-alfresco, line 46: Maybe instead we could have `List<String> contentPropertiesRemoved` here?

tpage-alfresco: Or maybe even better we could represent it internally as `CustomPropertyDeleted` in the `Set<CustomPropertyDelta<?>> properties` field?

Pawel-608: I like this idea much more, but I would still create separate command for that - it costs us basically nothing and will reduce the complexity

Pawel-608: we agreed to put content deletion info in `Set<CustomPropertyDelta<?>> properties`

krdabrowski (approved):
LGTM

PR #153, opened 2024-02-12
ACS-6685 Refactor users into the general properties.

tpage-alfresco:
Replace unit test for null users in log in event with e2e test, since we've removed the special code to handle this case.


krdabrowski (approved):
LGTM

PR #157, opened 2024-02-14
Fix localstack so that it works on Linux.

tpage-alfresco:


distribution/src/main/resources/docker-compose/docker-compose.yml
```diff
@@ -305,6 +305,8 @@ services:
       - "4566:4566"
     volumes:
       - './init-aws.sh:/etc/localstack/init/ready.d/init-aws.sh'
+      - "${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack"
```
krdabrowski, line 308: Single ⚔️ double quotes - I'm OK with the second option, but to be consistent in it you should fix also the above line

krdabrowski (approved):
I have only one small remark - overall nice fix! I've verified it and it works fine on my machine

distribution/src/main/resources/docker-compose/docker-compose.yml
tpage-alfresco: Ok :+1: Let's use double quotes since `init-aws.sh` is the only line in the whole file with single quotes.

PR #161, opened 2024-02-15
ACS-6649: [HxI LI] Authentication to HxI API

krdabrowski:
https://alfresco.atlassian.net/browse/ACS-6649


live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/auth/AuthenticationService.java
```diff
@@ -0,0 +1,177 @@
...
+@Service
+@RequiredArgsConstructor
+public class AuthenticationService
+{
+    private static final String CLIENT_REGISTRATION_ID = "hyland-experience-auth";
+    private static final String APP_NAME_ATTRIBUTE_KEY = "applicationName";
+    private static final String SERVICE_USER_ATTRIBUTE_KEY = "serviceUser";
+    private static final String ENVIRONMENT_KEY_ATTRIBUTE_KEY = "hxAiEnvironmentKey";
+    private static final String ENVIRONMENT_KEY_HEADER = "hxai-environment";
+    private static final int AUTH_SCHEDULE_DELAY_MINUTES = 55;
```
krdabrowski, line 68: This could be a property in app.yaml configuration 🤔

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/auth/AuthenticationService.java
```diff
@@ -0,0 +1,177 @@
...
+    private static final int WAIT_FOR_PAUSE_TIME_MILLIS = 100;
```
tpage-alfresco, line 69: It's probably worth moving some of these values to the configuration file (it could be in another PR though).

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/auth/AuthenticationService.java
```diff
@@ -0,0 +1,177 @@
...
+
+    private final OAuth2ClientProperties oAuth2ClientProperties;
+    private final IntegrationProperties integrationProperties;
+    private final AuthenticationManager authenticationManager;
+    private final TaskScheduler taskScheduler;
+    private final CamelContext camelContext;
+
+    // Temporary disabled
+    // @PostConstruct
```
tpage-alfresco, line 78: Is this temporarily disabled for this PR and it will be enabled in a future one?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/auth/AuthenticationService.java
```diff
@@ -0,0 +1,177 @@
...
+        };
+        delegatingTaskScheduler.scheduleWithFixedDelay(authenticationTask, Duration.ofMinutes(AUTH_SCHEDULE_DELAY_MINUTES));
+    }
+
+    public void authenticate()
+    {
+        authenticate(false);
+    }
+
+    public void authenticate(boolean forceAuthentication)
```
tpage-alfresco, line 95: This doesn't ever seem to be set to anything but `false`.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/auth/AuthenticationService.java
```diff
@@ -0,0 +1,177 @@
...
+    private boolean tokenHasOrIsAboutToExpire(OAuth2AccessToken accessToken, int scheduleDelayMinutes)
+    {
+        if (Optional.ofNullable(accessToken.getIssuedAt()).isEmpty() || Optional.ofNullable(accessToken.getExpiresAt()).isEmpty())
+        {
+            return true;
+        }
+
+        Duration expirationOffset = Duration.between(accessToken.getIssuedAt(), accessToken.getExpiresAt()).minusMinutes(scheduleDelayMinutes);
+        long expirationOffsetSeconds = expirationOffset.isNegative() ? 0 : expirationOffset.toSeconds();
+        return Instant.now().isAfter(accessToken.getExpiresAt().minusSeconds(expirationOffsetSeconds));
```
tpage-alfresco, line 148: I didn't understand the code that is meant to handle "or is about to expire". I didn't think we would need something like this anyway because we planned to use a period of 55 minutes for a 60 minute token.

tpage-alfresco: Ah - is the idea that you're checking if the expiry will be before the next execution of this task?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/auth/AuthenticationService.java
krdabrowski: This is a leftover after testing 🤔  You are right, that this is not required at this moment - I will remove it

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/auth/AuthenticationService.java
krdabrowski: Yes, right now this is preventing us from wall of errors at startup due to unreachable auth URL

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/auth/AuthenticationService.java
krdabrowski: I was thinking about moving this delay to configuration file - on it right now

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/auth/AuthenticationService.java
krdabrowski: While testing I had a lot of calls earlier than 55 minutes form token creation and this check was preventing the app from generating a new one when it wasn't necessary 🤔 I'm not sure if it will be useful though, so I can remove it

krdabrowski: I'm thinking about running this logic every 55 min or in case of "access denied exception", or 401 🤔

tpage-alfresco: Yes - I assumed we would try to authenticate every 55 minutes and that way we'd ensure we always had at least one (sometimes two) tokens that would work.

tpage-alfresco (approved):
Nice work Krys!

PR #171, opened 2024-02-21
Gathering default tags in one place

krdabrowski:


live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/util/DockerTags.java
```diff
@@ -42,6 +42,9 @@
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public class DockerTags
 {
+    public static final String ACTIVEMQ_TAG_DEFAULT = "5.18.3-jre17-rockylinux8";
+    public static final String WIREMOCK_TAG_DEFAULT = "3.4.1";
+    public static final String LOCALSTACK_TAG_DEFAULT = "3.1.0";
```
tpage-alfresco, line 47: If we're putting these here then how about we make them private and instead add helper methods `getActiveMQTag`, `getWiremockTag`, `getLocalStackTag`?

tpage-alfresco (approved):
Definitely an improvement :+1:

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/util/DockerTags.java
krdabrowski: Good hint - I will fix it

PR #177, opened 2024-02-22
ACS-6750: create script that will generate custom namespace-prefixes.json

Pawel-608:


scripts/utils/namespaces-to-namespace-prefixes-file-generator.py
```diff
@@ -0,0 +1,127 @@
...
+PAGE_SIZE = 100
+ENDPOINT = "/alfresco/api/-default-/public/alfresco/versions/1/types"
+
+
+def is_input_valid():
+    return len(sys.argv) == 4 and not (None in sys.argv)
+
+
+def should_print_help():
+    return not is_input_valid() or sys.argv[1] == "help" or sys.argv[1] == "-h" or sys.argv[1] == "--help"
```
tpage-alfresco, line 39: I would use `argparse` for this. Even if it feels a little overkill then it makes the code simpler.

Pawel-608: meh I wanted to keep it super-simple and do not add this lib - idk if it'll add any readability. I cleaned up console args part - can you take a look one more time?

tpage-alfresco: Ok - let's look at this separately then. There's no fault in this PR anyway.

PR #186, opened 2024-02-26
Fix unit tests folder structure

Pawel-608:


Pawel-608:
For some reason we had some tests in `java/live_ingester` directory instead of `java/org/alfresco/hx_insight/live_ingester` - so I moved all of them to correct location

krdabrowski (approved):
LGTM

PR #191, opened 2024-02-27
ACS-6389: [HxI PH] Listen to predictions from HxI

krdabrowski:
https://alfresco.atlassian.net/browse/ACS-6389


prediction-applier/src/integration-test/java/org/alfresco/hxi_connector/prediction_applier/util/DockerTags.java
```diff
@@ -0,0 +1,124 @@
...
+import lombok.AccessLevel;
+import lombok.Cleanup;
+import lombok.NoArgsConstructor;
+import lombok.SneakyThrows;
+
+/**
+ * Allows to access properties from 'docker-tags.properties' file. It's required to run: `mvn package` before accessing the properties file, to allow Maven filter out and replace variables.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class DockerTags
```
mpichura, line 43: This will be the third class named `DockerTags` in the project. From what I can see, there are at least several methods which seem to be either exactly the same or have same purpose.
Have we thought about creating a common-codebase module inside the project? That could also be of use for node filtering.

mpichura (approved):
LGTM. 
While looking at the code I had some idea about common codebase. We may want to discuss that teamwise.

prediction-applier/src/integration-test/java/org/alfresco/hxi_connector/prediction_applier/util/DockerTags.java
tpage-alfresco: You're probably right. We did think about it before, but there wasn't much duplication. Now I can think of several things:
* `DockerTags` and other Testcontainers code
* Constant for names of properties like `cm:content`
* Prediction model
* Bulk ingestion -> Live ingestion message format? (I don't remember seeing duplication of code here, but I think there must be)

distribution/src/main/resources/docker-compose/docker-compose.yml
```diff
@@ -319,8 +332,9 @@ services:
         limits:
           memory: 256m
     environment:
-      SERVICES: s3
+      SERVICES: s3,sqs
```
tpage-alfresco, line 335: I don't think the messages are going to come via SQS. It's not been defined yet, but we might need to poll on a REST API. I have no problem with this being the first version of our code though.

prediction-applier/src/integration-test/java/org/alfresco/hxi_connector/prediction_applier/util/DockerTags.java
krdabrowski: I was thinking the same while copying this class, but lets do it in separate PR

distribution/src/main/resources/docker-compose/docker-compose.yml
krdabrowski: I'm aware of that, but for now lets leave it as it is, as there's not much we can relay on

PR #197, opened 2024-03-05
ACS-6929 Insert "value" before all property values.

tpage-alfresco:
Note that this is technically wrong for content, but we will fix that in a future PR.


mpichura (approved):
LGTM

PR #202, opened 2024-03-06
ACS-6871: Adding Node Type filtering for Repo Events.

mpichura:


live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
```diff
@@ -0,0 +1,67 @@
...
+@Slf4j
+public class TypeFilterApplier implements NodeFilterApplier
+{
+    @Override
+    public boolean applyFilter(RepoEvent<DataAttributes<NodeResource>> repoEvent, Filter filter)
+    {
+        final String nodeType = repoEvent.getData().getResource().getNodeType();
+        final List<String> allowed = filter.type().allow();
+        final List<String> denied = filter.type().deny();
+        log.atDebug().log("Applying type filters on repo event of id: {}. Event node type: {}. Allowed types: {}. Denied types: {}", repoEvent.getId(), nodeType, allowed, denied);
```
Pawel-608, line 52: why not `log.debug()`?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    private boolean filterAllowed(String nodeType, List<String> allowed)
+    {
+        return allowed.stream().anyMatch(a -> a.equals(nodeType)) || CollectionUtils.isEmpty(allowed);
+    }
+
+    private boolean filterDenied(String nodeType, List<String> denied)
+    {
+        return !CollectionUtils.isEmpty(denied) && denied.stream().anyMatch(d -> d.equals(nodeType));
```
Pawel-608, line 65: maybe `isAllowed` and `isDenied` instead?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
```diff
@@ -0,0 +1,67 @@
...
+{
+    @Override
+    public boolean applyFilter(RepoEvent<DataAttributes<NodeResource>> repoEvent, Filter filter)
+    {
+        final String nodeType = repoEvent.getData().getResource().getNodeType();
+        final List<String> allowed = filter.type().allow();
+        final List<String> denied = filter.type().deny();
+        log.atDebug().log("Applying type filters on repo event of id: {}. Event node type: {}. Allowed types: {}. Denied types: {}", repoEvent.getId(), nodeType, allowed, denied);
+        final boolean allow = filterAllowed(nodeType, allowed);
+        final boolean deny = filterDenied(nodeType, denied);
```
Pawel-608, line 54: I would get rid of these `final` keywords, imo they are unnecessary

Pawel-608 (approved):
LGTM, just small remarks

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
mpichura: It's a shortcut to wrap the logging around a check if debug log level is enabled (log guarding). It's (presumably) more performant when you use parametrized logging, because you don't consume resources of `toString` and other methods (if used in retrieving the parameters).

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
mpichura: These are not `boolean` values - they are lists of denied/allowed values.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
mpichura: But these variables are `final`... I don't get why wouldn't I "mark" them final then...

krdabrowski (approved):
LGTM

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
Pawel-608: hmm, but the goal of the method is to say if node with some type is allowed to pass the filter or not (it does not perform any filtering operation as the name suggests)

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
Pawel-608: idk it is just an opinion but for me it makes unnecessary noise - you can clearly see that you dont reassign any of these variables (even ide will highlight variables that are reassigned)

mpichura: I've been against marking all variables/method params `final` for a long time especially that at some point IntelliJ's `Extract to variable` started to do that out of the box (it lasted quite for short time). Then I got used to it and started seeing more values in having a habit to that for each variable/param. It sometimes can make people think twice when you try to re-assign and then your IDE goes red - without `final` it doesn't. And believe some people only start re-thinking when IDE goes red.
From my experience it does no harm but can do good, so sorry - but we seem to have different opinions.

Pawel-608: yep but at the end of the day, it is all just about consistency - we should always add or always skip it ;) Closing the conversation but imo it is something that is worth to standardise

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
```diff
@@ -0,0 +1,67 @@
+/*
+ * #%L
+ * Alfresco HX Insight Connector
+ * %%
+ * Copyright (C) 2023 Alfresco Software Limited
```
tpage-alfresco, line 5: This is something we should automate. I know we discussed this in the past, but how about if headers were automatically updated for files modified in a commit by pre-commit? This seems uncontroversial and would happen in the same way as our auto-formatting does.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
```diff
@@ -0,0 +1,67 @@
...
+        final List<String> denied = filter.type().deny();
+        log.atDebug().log("Applying type filters on repo event of id: {}. Event node type: {}. Allowed types: {}. Denied types: {}", repoEvent.getId(), nodeType, allowed, denied);
+        final boolean allow = filterAllowed(nodeType, allowed);
+        final boolean deny = filterDenied(nodeType, denied);
+        return allow && !deny;
+    }
+
+    private boolean filterAllowed(String nodeType, List<String> allowed)
+    {
+        return allowed.stream().anyMatch(a -> a.equals(nodeType)) || CollectionUtils.isEmpty(allowed);
```
tpage-alfresco, line 60: How about:
```suggestion
        return CollectionUtils.isEmpty(allowed) || allowed.contains(nodeType);
```

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
tpage-alfresco: If we know that `denied` isn't null then I think this can be:
```suggestion
        return denied.contains(nodeType);
```

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplierTest.java
```diff
@@ -0,0 +1,168 @@
...
+
+        // then
+        assertTrue(result);
+    }
+
+    @Test
+    void shouldNotFilterOutWhenEmptyAllowedAndTypeNotInDenied()
+    {
+        final String type1 = "cm:folder";
+        final String type2 = "cm:special-folder";
```
tpage-alfresco, line 109: You can make these constants to remove some duplication.

tpage-alfresco (approved):
Code looks good :+1: I had some suggestions to try to simplify the filter logic.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
mpichura: I have a spotless fix for that - will introduce in a separate PR.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplierTest.java
mpichura: I was thinking about that here and in the aspect filter test class. Will do.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
mpichura: Thanks, I'll soon change this.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplier.java
mpichura: > hmm, but the goal of the method is to say if node with some type is allowed to pass the filter or not (it does not perform any filtering operation as the name suggests)

I'll change method names - I thought you were thinking about parameter names.

PR #204, opened 2024-03-07
ACS-6869: Fixing allow/deny logic, license headers and tests.

mpichura:


tpage-alfresco (approved):
LGTM!

Pawel-608 (approved):
LGTM (maybe it is just worth to add one test that will cover aspectNames=null sorry if I missed it)

mpichura:
> LGTM (maybe it is just worth to add one test that will cover aspectNames=null sorry if I missed it)

The integration test I added here has `"aspectNames": [ ],` but I guess you had in mind a unit test with `null` aspects coming from `repoEvent`, right? I'll ad one - good point.

krdabrowski (approved):

PR #207, opened 2024-03-08
NO_TICKET: Adding spotless license header config

mpichura:
This config will introduce 2 changes in spotless apply/check:
1. Whole run will be executed against only the files which differ from `origin/master` branch.
2. All files will be updated with header from `license-header` file in root of the project. If the file doesn't have any header - it will just be added, If there already is a header, then if the year is not current year it will be updated to `<previous_year_value>-<current_year>`, e.g., from `2023` to `2023-2024`.
For details see https://github.com/diffplug/spotless/tree/main/plugin-maven#license-header and https://github.com/diffplug/spotless/tree/main/plugin-maven#ratchet

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -349,6 +349,11 @@
                             <includes>
                                 <include>src/*/java/**/*.java</include>
                             </includes>
+                            <ratchetFrom>origin/master</ratchetFrom> <!-- this means that the whole spottless check/apply will run only against differences to origin/master branch-->
+                            <licenseHeader>
+                                <file>license-header</file>
+                                <delimiter>package</delimiter>  <!-- content until first occurrence of the delimiter regex will be interpreted as header section -->
+                            </licenseHeader>
                             <eclipse>
                                 <file>alfresco-formatter.xml</file>
                             </eclipse>
--- a/pom.xml
+++ b/pom.xml
@@ -349,7 +349,7 @@
                             <includes>
                                 <include>src/*/java/**/*.java</include>
                             </includes>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -349,6 +349,11 @@
                             <includes>
                                 <include>src/*/java/**/*.java</include>
                             </includes>
+                            <ratchetFrom>origin/master</ratchetFrom> <!-- this means that the whole spottless check/apply will run only against differences to origin/master branch-->
```
tpage-alfresco, line 352: I don't think we want this as the comparison point. We want the changes going into the current commit.

tpage-alfresco (commented):
I think this solution is a lot neater codewise than what I ended up with from the maven license plugin. I imagine it's quicker overall too (because we're already running Spotless against all java files).

I don't like the license header being stored in this repository. It's already defined as a maven dependency for all projects.

Additionally I think we can't compare against `origin/master`, but I think that shouldn't be too hard to fix.

mpichura:
> Additionally I think we can't compare against `origin/master`, but I think that shouldn't be too hard to fix.
This, surprisingly, can be quite hard to achieve to satisfy local development and pull request needs. I couldn't come up with a decent way to point to a revision that would work all way. 
Thus, I'll also try with https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#incremental-up-to-date-checking-and-formatting

mpichura:
We decided to go with https://github.com/Alfresco/hxinsight-connector/pull/208

PR #208, opened 2024-03-08
ACS-6952 Automatically update modified license headers using maven plugin.

tpage-alfresco:


tpage-alfresco (commented):
The code in the hook is awkward because there doesn't seem to be a way to pass a list of paths into the license plugin. It has to update all headers and then revert.

This is slower than the solution in #207 because we're currently not running maven at all with the existing pre-commit (whereas we are running spotless).

As a side note the headers get updated to have two dates - e.g. `2005-2024`. I guess this is the format specified in the original license definition. I believe we can override the 2005 with the `inceptionYear` option if we want.

.pre-commit-config.yaml
```diff
@@ -14,11 +13,21 @@ repos:
...
+        entry: ./scripts/hooks/spotless-check-and-format.sh
+        language: script
+        files: ".*.java"
+        pass_filenames: false
+      - id: license-header-check-and-update
+        name: License header check and update
+        entry: ./scripts/hooks/license-check-and-update.sh
+        language: script
+        exclude: ""
+        always_run: true
```
Pawel-608, line 29: not sure if it was intentional but if you want to run this script only once you should add `pass_filenames: false` - otherwise it will be invoked one time for every single file

tpage-alfresco: Specifying `exclude` achieves the same thing. You're right that their documentation suggests using `pass_filenames: false` but honestly I think that's less clear - I expected it to run against all files but without passing filenames. I don't mind changing `exclude: ""` to `pass_filenames: false` though - they're both pretty unclear! :joy:

tpage-alfresco: Actually - I just saw we have `pass_filenames: false` above, and I planned to merge the two steps (to half the build time by only invoking maven once). I'll merge the two steps and use the existing definition.

mpichura (approved):
Nice job.
I guess I'll close https://github.com/Alfresco/hxinsight-connector/pull/207 now, WDYT?

tpage-alfresco:
I think it's fine to close #207 - I think this does everything I wanted. It feels like there might be a neater solution by using spotless for both checks, but maybe not since the license headers are stored in Nexus.

krdabrowski (approved):

PR #217, opened 2024-03-13
ACS-6869: Fixing filter logic.

mpichura:
This PR fixes aspect filtering logic for a case when not all aspects present on a node are present in allow list (but some are).
That seemed to be a problem with integration test (as the unit test tested the wrong logic but integration test was testing the correct one).
Edit:
I reverted filtering logic. Only changes now are test fixes.

### DIFF
--- a/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/AspectFilterApplier.java
+++ b/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/AspectFilterApplier.java
@@ -59,18 +59,7 @@ public class AspectFilterApplier implements NodeFilterApplier
 
     private boolean isAllowed(Set<String> aspectNames, List<String> allowed)
     {
-        if (allowed.isEmpty())
-        {
-            return true;
-        }
-        else if (aspectNames.isEmpty())
-        {
-            return false;
-        }
-        else
-        {
-            return new HashSet<>(allowed).containsAll(aspectNames);
-        }
+        return allowed.isEmpty() || !aspectNames.isEmpty() && new HashSet<>(allowed).containsAll(aspectNames);
     }
 
     private boolean isDenied(Set<String> aspectNames, List<String> denied)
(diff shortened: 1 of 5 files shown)


krdabrowski (approved):
Nice fix!

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/AspectFilterApplier.java
```diff
@@ -59,7 +59,7 @@ public boolean applyFilter(RepoEvent<DataAttributes<NodeResource>> repoEvent, Fi
 
     private boolean isAllowed(Set<String> aspectNames, List<String> allowed)
     {
-        return CollectionUtils.isEmpty(allowed) || allowed.stream().anyMatch(aspectNames::contains);
+        return allowed.isEmpty() || !aspectNames.isEmpty() && new HashSet<>(allowed).containsAll(aspectNames);
```
tpage-alfresco, line 62: I don't think this is correct:
```
aspectNames: a, b
allowed: a, c
allowed.containsAll(aspectNames) -> false
```
In this case I'd expect the presences of aspect `a` to qualify the node.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/AspectFilterApplierTest.java
```diff
@@ -160,7 +160,7 @@ void shouldNotFilterOutWhenAspectInAllowedAndEmptyDenied()
         boolean result = objectUnderTest.applyFilter(mockRepoEvent, mockFilter);
 
         // then
-        assertTrue(result);
+        assertFalse(result);
```
tpage-alfresco, line 163: I think this should be `assertTrue`.

tpage-alfresco (changes requested):
I think the original logic was correct. Perhaps the integration test needs updating?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/AspectFilterApplier.java
mpichura: Reverted

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/AspectFilterApplierTest.java
mpichura: Reverted

tpage-alfresco (approved):
LGTM!

PR #227, opened 2024-03-15
ACS-6874: Moving filters logic to common module.

mpichura:
Moving filtering logic to common module as it will be used in bulk ingester.

### DIFF
--- a/common/src/test/java/org/alfresco/hxi_connector/common/repository/filter/AspectFilterTest.java
+++ b/common/src/test/java/org/alfresco/hxi_connector/common/repository/filter/AspectFilterTest.java
@@ -70,22 +70,22 @@ class AspectFilterTest
     private static Stream<Arguments> provideParameters()
     {
         return Stream.of(
-                composeArguments(Set.of(CM_ASPECT_1), emptyList(), emptyList(), true),
-                composeArguments(emptySet(), List.of(CM_ASPECT_1), emptyList(), false),
-                composeArguments(emptySet(), emptyList(), emptyList(), true),
-                composeArguments(Set.of(CM_ASPECT_1, CM_ASPECT_2), List.of(CM_ASPECT_1), emptyList(), true),
-                composeArguments(Set.of(CM_ASPECT_1), emptyList(), List.of(CM_ASPECT_2), true),
-                composeArguments(Set.of(CM_ASPECT_1), List.of(CM_ASPECT_1), List.of(CM_ASPECT_2), true),
-                composeArguments(Set.of(CM_ASPECT_1, CM_ASPECT_2), List.of(CM_ASPECT_1), List.of(CM_ASPECT_2), false),
-                composeArguments(Set.of(CM_ASPECT_1, CM_ASPECT_2), List.of(CM_ASPECT_3), emptyList(), false),
-                composeArguments(Set.of(CM_ASPECT_2, CM_ASPECT_3), emptyList(), List.of(CM_ASPECT_3), false),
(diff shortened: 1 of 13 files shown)


live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/AspectFilterApplierTest.java
```diff
@@ -79,34 +72,6 @@ void mockBasicData()
         given(mockFilter.aspect()).willReturn(mockAspect);
     }
 
-    @Test
-    void shouldNotFilterOutNonEmptyAspectsWhenEmptyAllowedAndEmptyDenied()
```
mpichura, line 83: All removed tests are reflected in `AspectFilterTest` class in common module. 
The ones that remained here are testing the only logic left in tested class which is mapping `null` aspects into empty set.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/TypeFilterApplierTest.java
```diff
@@ -76,86 +73,14 @@ void mockBasicData()
     }
 
     @Test
-    void shouldNotFilterOutWhenEmptyAllowedAndEmptyDenied()
+    void shouldThrowExceptionWhenNullNodeType()
```
mpichura, line 76: All removed tests are reflected in `TypeFilterTest` class in common module. 
This sole test is sort of a smoke test that checks if `null` node type is handled correctly (exception thrown).

common/src/test/java/org/alfresco/hxi_connector/common/repository/filter/AspectFilterTest.java
```diff
@@ -0,0 +1,91 @@
...
+        else
+        {
+            assertFalse(result);
+        }
+    }
+
+    private static Stream<Arguments> provideParameters()
+    {
+        return Stream.of(
+                composeArguments(Set.of(CM_ASPECT_1), emptyList(), emptyList(), true),
```
krdabrowski, line 73: Order of arguments is confusing form me, because it differs from what test method expects:
`void testAspectFiltering(boolean expected, Set<String> aspects, List<String> allowed, List<String> denied)`. IMO the order should be the same

common/src/test/java/org/alfresco/hxi_connector/common/repository/filter/TypeFilterTest.java
```diff
@@ -0,0 +1,87 @@
...
+        else
+        {
+            assertFalse(result);
+        }
+    }
+
+    private static Stream<Arguments> provideParameters()
+    {
+        return Stream.of(
+                composeArguments(CM_FOLDER, emptyList(), emptyList(), true),
```
krdabrowski, line 72: Here also order should match order of args in above method IMO

krdabrowski (approved):
Overall LGTM, I have only minor remarks

common/src/test/java/org/alfresco/hxi_connector/common/repository/filter/AspectFilterTest.java
mpichura: Good point - fixed that.

common/src/test/java/org/alfresco/hxi_connector/common/repository/filter/TypeFilterTest.java
mpichura: Good point - fixed that.

PR #230, opened 2024-03-18
ACS-6929 Content property should be labelled "file" not "value".

tpage-alfresco:


mpichura (approved):
LGTM

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/metadata/model/property/ContentPropertyUpdated.java
```diff
@@ -0,0 +1,86 @@
...
+        return Optional.empty();
+    }
+
+    public static ContentPropertyUpdatedBuilder builder(String propertyName)
+    {
+        return new ContentPropertyUpdatedBuilder(propertyName);
+    }
+
+    @RequiredArgsConstructor
+    public static class ContentPropertyUpdatedBuilder
```
krdabrowski, line 70: I wonder if you could use here Lombok's `@Builder` https://projectlombok.org/features/Builder 🤔

tpage-alfresco: I did look, but I didn't find a way to add the mandatory parameter to the builder constructor.

PR #231, opened 2024-03-18
ACS-6872: Adding in-memory Alfresco node aspect filtering.

mpichura:


krdabrowski (approved):
LGTM

bulk-ingester/src/main/resources/application.yml
```diff
@@ -1,23 +1,30 @@
...
+  datasource:
+    url: jdbc:postgresql://localhost:5432/alfresco
+    username: alfresco
+    password: alfresco
+    hikari.maximumPoolSize: 20
 
-alfresco.reindex.pathCacheSize: 10000
+alfresco:
+  reindex:
+    pathCacheSize: 10000
```
tpage-alfresco, line 11: The original single line property looked ok to me. Do you prefer presenting it across multiple lines?

tpage-alfresco: Ah - I see that the `alfresco` part was duplicated. So we could keep some of it compressed:
```
alfresco:
  reindex.pathCacheSize: 10000
  bulk.ingest:
    node-params:
      fromId: 0
      toId: 20000000000
  ...
```

bulk-ingester/src/integration-test/java/org/alfresco/hxi_connector/bulk_ingester/processor/BulkIngestionFilterIntegrationTest.java
```diff
@@ -0,0 +1,90 @@
...
+                .build();
+        IngestEvent folder = IngestEvent.builder()
+                .nodeId("dad275aa-affc-487d-a7ed-92cf8e6ce351")
+                .properties(parseProperties("type=cm:folder", "cm:name=Animals", "cm:title=Animals overview", "createdAt=1708329323", "aspectsNames=[cm:titled, cm:auditable]", "createdBy=admin", "cm:description=This folder contains overview of the animals in our zoo", "modifiedBy=admin"))
+                .build();
+        IngestEvent textFile = IngestEvent.builder()
+                .nodeId("44545a62-0f64-4d3e-838a-9f8ba23df0c7")
+                .contentInfo(new IngestEvent.ContentInfo(28, "UTF-8", "text/plain"))
+                .properties(parseProperties("cm:name=Animals list", "cm:title=List of animals", "type=cm:content", "createdAt=1708329628", "app:editInline=true", "aspectsNames=[cm:generalclassifiable, app:inlineeditable, cm:thumbnailModification, cm:titled, cm:taggable, rn:renditioned, cm:auditable]", "cm:categories=25805c3b-dc41-4a3d-9b03-a44ac7963c70", "createdBy=admin", "cm:lastThumbnailModification=[doclib:1708329633941, pdf:1708329629285]", "cm:description=List of animals in our zoo", "modifiedBy=admin"))
```
tpage-alfresco, line 72: I suggest removing `auditable` from one of these three nodes so that we test it's excluded.

bulk-ingester/src/integration-test/java/org/alfresco/hxi_connector/bulk_ingester/processor/DummyIngestEventPublisher.java
```diff
@@ -0,0 +1,76 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+import org.alfresco.hxi_connector.bulk_ingester.event.IngestEventPublisher;
+import org.alfresco.hxi_connector.common.model.ingest.IngestEvent;
+
+@Slf4j
+public class DummyIngestEventPublisher implements IngestEventPublisher
+{
+    private final Map<String, IngestEvent> ingestEvents = new HashMap<>();
```
tpage-alfresco, line 46: I think this is only used by one test, but it might be good to reset it between tests to try to avoid dirty data.

tpage-alfresco (approved):
A few comments - some of them about code that was just moved in this PR. Looks really good though :+1:

bulk-ingester/src/main/resources/application.yml
mpichura: One thing - it was duplicated. Second one - I think we shouldn't mix text-style properties with yaml-style properties.

bulk-ingester/src/integration-test/java/org/alfresco/hxi_connector/bulk_ingester/processor/DummyIngestEventPublisher.java
mpichura: After my changes this one is used by 2 classes - and thanks - I'll do a clean up (even between each method).

bulk-ingester/src/integration-test/java/org/alfresco/hxi_connector/bulk_ingester/processor/BulkIngestionFilterIntegrationTest.java
mpichura: I thought I could do something like that but it appears it's not that easy.
We are preloading alfresco database with a dump which is used by this and several more tests - I was lazy enough to . 
I'll try to find a node without any of `allowed` aspects and include that in the test.

bulk-ingester/src/integration-test/java/org/alfresco/hxi_connector/bulk_ingester/util/IngestEventPropertyParser.java
```diff
@@ -0,0 +1,88 @@
...
+        return Map.entry(key, parsePropertyValue(value));
+    }
+
+    private static Serializable parsePropertyValue(String value)
+    {
+        if (value.startsWith("["))
+        {
+            value = value.replace("[", "");
+            value = value.replace("]", "");
+            String[] valueSplit = value.split(", ");
```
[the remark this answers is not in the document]
mpichura: Good catch - I'll split with comma and strip leading spaces, sounds sensible?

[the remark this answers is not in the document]
tpage-alfresco: Yep - sounds reasonable.

bulk-ingester/src/main/resources/application.yml
tpage-alfresco: Ok - there's also `hikari.maximumPoolSize` above.

PR #236, opened 2024-03-19
ACS-6389: add integration test verifying hxi:model installation

krdabrowski:
https://alfresco.atlassian.net/browse/ACS-6389


prediction-applier-extension/src/integration-test/java/org/alfresco/hxi_connector/prediction_applier/util/client/AspectsClient.java
```diff
@@ -0,0 +1,115 @@
...
+public class AspectsClient
+{
+    private static final String ASPECTS_URL = "http://%s:%s/alfresco/api/-default-/public/alfresco/versions/1/aspects?maxItems=%s";
+    private static final String USER = "admin";
+    private static final String PASS = "admin";
+
+    private final ObjectMapper objectMapper = new ObjectMapper();
+    private final String host;
+    private final int port;
+    private final int timeout;
```
tpage-alfresco, line 61: Maybe call this `timeoutSeconds` or something?

common-test/src/main/java/org/alfresco/hxi_connector/common/test/util/DockerContainers.java
```diff
@@ -0,0 +1,151 @@
...
+    private static final String ACTIVE_MQ_IMAGE = "quay.io/alfresco/alfresco-activemq";
+    private static final String ACTIVE_MQ_TAG = DockerTags.getActiveMqTag();
+    private static final String DB_USER = "alfresco";
+    private static final String DB_PASS = "alfresco";
+    private static final String DB_NAME = "alfresco";
+    private static final String REPOSITORY_ALIAS = "repository";
+    private static final String POSTGRES_ALIAS = "postgres";
+    private static final String ACTIVE_MQ_ALIAS = "activemq";
+
+    public static GenericContainer<?> createExtendedRepositoryContainerWithin(Network network)
```
mpichura, line 66: Did you try with https://java.testcontainers.org/modules/docker_compose/ approach?
I am not saying it's better but it might make use of docker compose files we already have in `distribution` module. On the other hand, I am not sure how docker tags from generated `.env` file would be handled,

mpichura (approved):
Nice job. I had one suggestion on requesting single aspect rather than the whole list.

azakrzewski-hy (approved):

prediction-applier-extension/src/integration-test/java/org/alfresco/hxi_connector/prediction_applier/util/client/AspectsClient.java
```diff
@@ -0,0 +1,115 @@
...
+        Map<String, Map<String, Object>> response = objectMapper.readValue(responseBody, Map.class);
+
+        List<Map<String, Object>> entries = ((List<Map<String, Map<String, Object>>>) response.get("list").get("entries")).stream()
+                .map(e -> e.get("entry"))
+                .collect(Collectors.toList());
+
+        return objectMapper.convertValue(entries, new TypeReference<>() {});
+    }
+
+    private String requestAspects(int maxItems) throws IOException, AuthenticationException
```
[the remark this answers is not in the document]
krdabrowski: Very good hint! I've fixed the code - thanks!

common-test/src/main/java/org/alfresco/hxi_connector/common/test/util/DockerContainers.java
krdabrowski: I'm not in favor of using d-c file - it's not so flexible as programmatic containers - not all services are required while running tests. Also configuration of one service can vary for different tests. TBH I didn't tried it - that might work, but I expect it will introduce some extra issues on the way 🤔

PR #240, opened 2024-03-20
ACS-6874: Adding primary path/ancestor node filtering in live ingester.

mpichura:
I assumed that primary path/ancestors will be under same filtering rules as aspects, thus, I've made small generalization for filter utility classes.

### DIFF
--- a/common/src/test/java/org/alfresco/hxi_connector/common/repository/filter/CollectionFilterTest.java
+++ b/common/src/test/java/org/alfresco/hxi_connector/common/repository/filter/CollectionFilterTest.java
@@ -31,6 +31,7 @@ import static java.util.Collections.emptySet;
 import static org.junit.jupiter.api.Assertions.*;
 import static org.junit.jupiter.api.Named.named;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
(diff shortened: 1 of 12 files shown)


common/src/test/java/org/alfresco/hxi_connector/common/repository/filter/CollectionFilterTest.java
```diff
@@ -67,7 +72,22 @@ void testAspectFiltering(boolean expected, Set<String> aspects, List<String> all
         }
     }
 
-    private static Stream<Arguments> provideParameters()
+    @ParameterizedTest
+    @MethodSource("provideAncestorParameters")
+    void testAncestorFiltering(boolean expected, List<String> aspects, List<String> allowed, List<String> denied)
```
krdabrowski, line 77: I assume in args there should be `List<String> ancestors` instead of `List<String> aspects`

krdabrowski (approved):
Nice generalization!

common/src/test/java/org/alfresco/hxi_connector/common/repository/filter/CollectionFilterTest.java
mpichura: Thanks, fixed that.

common/src/test/java/org/alfresco/hxi_connector/common/repository/filter/CollectionFilterTest.java
```diff
@@ -67,7 +72,22 @@ void testAspectFiltering(boolean expected, Set<String> aspects, List<String> all
...
+    {
+        boolean result = CollectionFilter.filter(ancestors, allowed, denied);
+        if (expected)
+        {
+            assertTrue(result);
+        }
+        else
+        {
+            assertFalse(result);
+        }
```
tpage-alfresco, line 87:
```suggestion
        assertEquals("Unexpected result from filter", result, expected);
```

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/RequestFiltersIntegrationTest.java
```diff
@@ -37,13 +37,15 @@
         "alfresco.filter.aspect.allow[0]=cm:versionable", "alfresco.filter.aspect.allow[1]=cm:auditable",
         "alfresco.filter.type.deny[0]=cm:folder",
         "alfresco.filter.type.allow[0]=cm:content",
+        "alfresco.filter.path.allow[0]=5f355d16-f824-4173-bf4b-b1ec37ef5549", "alfresco.filter.path.allow[1]=93f7edf5-e4d8-4749-9b4c-e45097e2e19d",
+        "alfresco.filter.path.deny[0]=93f7edf5-e4d8-4749-9b4c-e45097e2e19e",
```
tpage-alfresco, line 41: Maybe edit these allow and deny UUIDs to be a bit more distinctive? We could use e.g. `11111111-1111-1111-1111-111111111111` as an easier to remember identifier?

tpage-alfresco (approved):
This is great! I added a couple of suggestions.

Do we get the primary hierarchy from the bulk ingester too? (I guess we should, since it's indexed in ESC).

mpichura:
> Do we get the primary hierarchy in events sent from the bulk ingester too? (I guess we should, since it's indexed in ESC).
Unfortunately, we only get primary parent. In ESC we traverse through all parents until we get to the root. I guess we'll have to do the same here.

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/RequestFiltersIntegrationTest.java
mpichura: Brilliant suggestion, will do.

tpage-alfresco:
Ah - interesting, so if customers need to rely on this filter then they cannot use it for bulk ingestion (currently).

I don't think that stops us from merging the feature, but it's a significant limitation. The other live ingestion filters will also work from bulk ingestion events.

common/src/test/java/org/alfresco/hxi_connector/common/repository/filter/CollectionFilterTest.java
mpichura: I thought about that when I was writing this assertion but somehow wanted to stick to `assertTrue`/`assrtFalse` approach with `boolean` assertion. But it does look much neater. I'll change that.

PR #246, opened 2024-03-22
ACS-7345 Hx Insight always expects a list of ingestion events.

tpage-alfresco:
The list can contain a single entry though.


mpichura (approved):
LGTM

PR #250, opened 2024-03-25
ACS-6951 Map filtered out update event to create/delete when needed

mpichura:
This may not be the neatest solution but I think it is optimal enough to not degrade performance as most of logic is included in the filtering logic.
1. Main live ingester `Camel` route (`LiveIngesterEventHandler`) is now handling only `Camel`'s `Exchange` object (that should make it easier to get rid of `Camel` in the future)
2. Filtering logic for update events can enhance `Exchange` object with a custom property indicating a `CREATE` or `DELETE` event type needs to be sent to HxI.
3. When above mentioned `Exchange` property is present, `CamelEventMapper` will use it to map event type from `Exchange` to `RepoEvent`.

### DIFF
--- a/live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/RequestFiltersIntegrationTest.java
+++ b/live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/RequestFiltersIntegrationTest.java
@@ -1287,4 +1287,75 @@ public class RequestFiltersIntegrationTest extends E2ETestBase
                 ]""";
         containerSupport.expectHxIngestMessageReceived(expectedBody);
     }
+
+    @Test
+    void testUpdateRequestWhenSingleFilterDeniedNodeBeforeAndAnotherSingleFilterDeniesCurrentNode()
+    {
+        // given
+        String repoEvent = """
+                {
+                  "specversion": "1.0",
+                  "type": "org.alfresco.event.node.Updated",
+                  "id": "ae5dac3c-25d0-438d-b148-2084d1ab0517",
+                  "primaryHierarchy": [ "11111111-1111-1111-1111-111111111111", "5f355d16-f824-4173-bf4b-b1ec37ef5549", "93f7edf5-e4d8-4749-9b4c-e45097e2e19d" ],
+                  "source": "/08d9b620-48de-4247-8f33-360988d3b19b",
+                  "time": "2021-01-26T10:29:42.99524Z",
+                  "dataschema": "https://api.alfresco.com/schema/event/repo/v1/nodeUpdated",
+                  "datacontenttype": "application/json",
+                  "data": {
+                    "eventGroupId": "b5b1ebfe-45fc-4f86-b71b-421996482817",
+                    "resource": {
+                      "@type": "NodeResource",
+                      "id": "d71dd823-82c7-477c-8490-04cb0e826e17",
+                      "name": "purchase-order-scan.pdf",
+                      "nodeType": "cm:folder",
+                      "createdByUser": {
+                        "id": "admin",
+                        "displayName": "Administrator"
+                      },
+                      "createdAt": "2024-03-02T11:14:15.695Z",
+                      "modifiedByUser": {
+                        "id": "abeecher",
+                        "displayName": "Alice Beecher"
+                      },
+                      "modifiedAt": "2024-03-06T10:29:42.529Z",
+                      "content": {
+                        "mimeType": "application/pdf",
+                        "sizeInBytes": 531152,
+                        "encoding": "UTF-8"
+                      },
+                      "properties": {
+                        "cm:title": "Purchase Order",
+                        "cm:versionType": null,
+                        "cm:versionLabel": "1.0",
+                        "cm:taggable": null
+                      },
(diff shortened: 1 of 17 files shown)


tpage-alfresco (commented):
How about a test for when the node goes from denied by one filter, to being denied by another?

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/RequestFiltersIntegrationTest.java
```diff
@@ -893,4 +893,398 @@ void testUpdateRequestWithAspectInAllowedFilterAndTypeInDeniedFilterAndAncestorI
...
+                        "displayName": "Administrator"
+                      },
+                      "properties": {
+                        "cm:title": null,
+                        "cm:versionType": "MAJOR",
+                        "cm:versionLabel": "1.0",
+                        "cm:taggable": null,
+                        "cm:description": "Old Description"
+                      },
+                      "aspectNames": [ "cm:versionable", "sc:secured" ]
```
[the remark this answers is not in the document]
mpichura: Yeah - that's kind of double edged sword... 
I was even thinking about creating a method which would be some kind of shortcut to `RepoEvent.builder().build()` with some default values but am not sure if that is a huge value...
Honestly - there probably is no perfect solution here...

mpichura:
> How about a test for when the node goes from denied by one filter, to being denied by another?

Good idea - I'll work on that.

mpichura:
> > How about a test for when the node goes from denied by one filter, to being denied by another?
> 
> Good idea - I'll work on that.

Test added

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/RequestFiltersIntegrationTest.java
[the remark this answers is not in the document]
tpage-alfresco: I think I prefer seeing the JSON to having a builder (for the e2es), but some data like "cm:title", "cm:versionType", etc. could be removed to make the input and output shorter.

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/RequestFiltersIntegrationTest.java
```diff
@@ -893,4 +893,469 @@ void testUpdateRequestWithAspectInAllowedFilterAndTypeInDeniedFilterAndAncestorI
...
+    @Test
+    void testUpdateRequestWhenSingleFilterDeniedNodeBeforeAndAnotherSingleFilterDeniesCurrentNode()
+    {
+        // given
+        String repoEvent = """
+                {
+                  "specversion": "1.0",
+                  "type": "org.alfresco.event.node.Updated",
+                  "id": "ae5dac3c-25d0-438d-b148-2084d1ab0517",
+                  "primaryHierarchy": [ "11111111-1111-1111-1111-111111111111", "5f355d16-f824-4173-bf4b-b1ec37ef5549", "93f7edf5-e4d8-4749-9b4c-e45097e2e19d" ],
```
tpage-alfresco, line 1300: This is in the wrong place. I can't work out what two filter properties changed for this test - the aspects and something else?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/utils/ExchangeEnricher.java
```diff
@@ -0,0 +1,69 @@
...
+     * This method will enrich Camel exchange with a property indicating that original event type should be changed to the value of that property. It is called during node filtering as most of the needed logic is already there.
+     *
+     * @param exchange
+     *            Camel exchange object
+     * @param resultBefore
+     *            Filtering result for node resource before
+     * @param result
+     *            Filtering result for current node resource
+     */
+    public static void enrichExchangeAfterFiltering(Exchange exchange, boolean resultBefore, boolean result)
```
tpage-alfresco, line 56: I think this method should fail if:
* Aspect filter changes to denied (sets UPDATED_EVENT_TYPE_PROP)
* Path filter changes to approved (overwrites UPDATED_EVENT_TYPE_PROP)

The end result should be that there is no event sent to Hx Insight (since the node was denied before and after).

tpage-alfresco: Is it possible to take the overall filter result and then do something like this, rather than calling this three times?

mpichura:
I am working on a bigger refactor of this feature which I'll present with a PR from a separate branch.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/utils/ExchangeEnricher.java
mpichura: I'm working out a neater (and reliable) way to call this once without option to overwrite.

mpichura:
Superseded by https://github.com/Alfresco/hxinsight-connector/pull/259

PR #254, opened 2024-03-26
ACS-6474: handle post-transformation errors

Pawel-608:


tpage-alfresco (commented):
This looks good but we should also have at least one test.

Pawel-608:
hmm, I thought I'll add tests in separate pr, but probably I don't really need full environment to test it - I'll try to add some tests in this pr

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/ATSTransformResponseHandlerTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+        // then
+        then(ingestContentCommandHandler).should(times(1)).handle(expectedCommand);
+        then(atsTransformRequester).shouldHaveNoInteractions();
+    }
+
+    @SneakyThrows
+    private void simulateResponse(TransformResponse response)
+    {
+        ObjectMapper objectMapper = new ObjectMapper();
```
tpage-alfresco, line 215: Can this be a field and then the whole method be replaced by the single call to camel?

In fact do we need to create a new `camelContext.createProducerTemplate()` for each test?

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/ATSTransformResponseHandlerTest.java
```diff
@@ -0,0 +1,244 @@
...
+        try
+        {
+            simulateResponse(transformResponse);
+        }
+        catch (CamelExecutionException ignored)
+        {}
+
+        // then
+        then(ingestContentCommandHandler).should(times(1)).handle(expectedCommand);
+        then(atsTransformRequester).should().requestTransformRetry(expectedTransformRequest, retryAttempt);
```
tpage-alfresco, line 174: You could put this in the catch block and have a `fail` statement if no exception was received.

tpage-alfresco (approved):
Thanks for adding some tests :+1:

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/ATSTransformResponseHandlerTest.java
Pawel-608: tbh I like current version more as it keeps a lot of test implementation details in this method and we don't have configuration mess at the start of the class (+ it costs us nothing in test, it does not have to be super performant and reuse everyting)

Speaking about ProducerTemplate - I have the same feelings as about ObjectMapper but I think it is also worth to mention that I used it in exactly same way we use it in production code (we are creating new ProducerTemplate for each request - maybe it is a mistake and we should create one and reuse it?)
image

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/ATSTransformResponseHandlerTest.java
Pawel-608: in real scenario this error is not thrown as I believe onException() is catching it and I'm not sure why it is thrown here so I would leave it as it is

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/request/ATSTransformRequester.java
```diff
@@ -96,9 +104,9 @@ private Map<String, String> getTransformRequestOptions(Transform transformProper
         return Map.of(TIMEOUT_KEY, String.valueOf(transformProperties.request().timeout()));
     }
 
-    private String makeClientDataString(TransformRequest transformRequest)
+    private String makeClientDataString(TransformRequest transformRequest, int attempt)
```
krdabrowski, line 107: At first glance I've red "mockClientDataString" and was confused, and started even wondering if is it a test 😂

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/ATSTransformResponseHandler.java
```diff
@@ -51,33 +55,66 @@ public class ATSTransformResponseHandler extends RouteBuilder
...
         ingestContentCommandHandler.handle(command);
     }
 
+    @SuppressWarnings("PMD.UnusedPrivateMethod")
+    private void retryContentIngestion(Exchange exchange)
+    {
+        Exception exception = exchange.getProperty(Exchange.EXCEPTION_CAUGHT, Exception.class);
+        TransformResponse transformResponse = exchange.getIn().getBody(TransformResponse.class);
+
+        int maxAttempts = integrationProperties.alfresco().transform().response().retryTransformation().attempts();
```
krdabrowski, line 105: This is transformation or ingestion retry? Method name indicates that ingestion but body is about transformation 🤔

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/ATSTransformResponseHandler.java
```diff
@@ -51,33 +55,66 @@ public class ATSTransformResponseHandler extends RouteBuilder
...
+    private final ATSTransformRequester atsTransformRequester;
 
     @Override
     public void configure()
     {
-        // @formatter:off
+        onException(EmptyRenditionException.class, ResourceNotFoundException.class)
+                .process(this::retryContentIngestion);
+
+        onException(Exception.class)
```
krdabrowski, line 66: Is retry triggered in case of every exception? This is expected?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/request/ATSTransformRequester.java
```diff
@@ -96,9 +104,9 @@ private Map<String, String> getTransformRequestOptions(Transform transformProper
...
     {
-        ClientData clientData = new ClientData(transformRequest.nodeRef(), transformRequest.targetMimeType());
+        ClientData clientData = new ClientData(transformRequest.nodeRef(), transformRequest.targetMimeType(), attempt);
         try
         {
             return objectMapper.writeValueAsString(clientData);
```
krdabrowski, line 112: Is it required here, after you added `ClientDataSerializer`? 🤔

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/ATSTransformResponseHandlerTest.java
```diff
@@ -0,0 +1,244 @@
...
+                nodeId);
+
+        doThrow(new LiveIngesterRuntimeException("Some exception")).when(ingestContentCommandHandler).handle(expectedCommand);
+
+        // when
+        try
+        {
+            simulateResponse(transformResponse);
+        }
+        catch (CamelExecutionException ignored)
```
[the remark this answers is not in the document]
Pawel-608: good idea! changed

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/request/ATSTransformRequester.java
Pawel-608: I think I we get rid of this method... I'll check ;)

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/ATSTransformResponseHandler.java
Pawel-608: good point - it is transformation retry, I changed method name

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/ATSTransformResponseHandler.java
Pawel-608: hah, there are 2 tickets and one says ([ACS-6474](https://hyland.atlassian.net/browse/ACS-6474)) that we should retry on any IOException (which is never thrown as it checked exception so I broaden it to just any Excpetion) and there is second ticket ([ACS-6475](https://hyland.atlassian.net/browse/ACS-6475)) that says that we shouldn't retry on any Exception😂 I would leave it as it is and create new ticket that will detail what should we handle in which way

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/request/ATSTransformRequester.java
Pawel-608: I actually can use it - I refactored this class a little

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/ATSTransformResponseHandlerTest.java
Pawel-608: I replaced CamelContext bean with ProducerTemplate bean

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/ATSTransformResponseHandler.java
krdabrowski: This sounds reasonable 👍

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/request/ATSTransformRequester.java
krdabrowski: Nice refactor!

krdabrowski (approved):
Good job!

PR #255, opened 2024-03-27
ACS-6951 alter filtered out update event to create delete

mpichura:
This is a refactor to https://github.com/Alfresco/hxinsight-connector/pull/250 (which will soon be closed).
Generally - there have been following changes made:
1. Unmarshalling of `RepoEvent` is done in main live ingester route (`LiveIngesterEventHandler`).
3. Filter handling is divided into following steps:
- determine filtering results for current and previous version of the node
- if node should be denied by filters, a custom property in Camel Exchange is created to indicate that
- if event type is `Updated` and filtering results have changed between current and previous version of the node, then event type is altered to `Created` or `Deleted`
4. Actual filtering is executed in main live ingester route basing on custom property in Camel Exchange existance.


live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/LiveIngesterEventHandler.java
```diff
@@ -54,7 +58,9 @@ public void configure()
                 .transacted()
                 .routeId(ROUTE_ID)
                 .log(DEBUG, "Received repo event : ${header.JMSMessageID}")
-                .filter(exchange -> repoEventFilterHandler.filterNode(exchange, integrationProperties.alfresco().filter()))
+                .setBody(camelEventMapper::repoEventFrom)
+                .process(exchange -> repoEventFilterHandler.handle(exchange, integrationProperties.alfresco().filter()))
+                .filter(exchange -> BooleanUtils.isNotTrue(exchange.getProperty(DENY_NODE, Boolean.class)))
```
tpage-alfresco, line 63: Could we just return a boolean from `repoEventFilterHandler.handle` and use that to filter (I'm aware I'm suggesting basically the code that used to be here).

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/AspectFilterApplier.java
```diff
@@ -27,43 +27,38 @@
...
-        return result;
+        log.atDebug().log("Applying aspect filters on node id: {}", nodeResource.getId());
+        return CollectionFilter.filter(aspectNames, allowed, denied);
+    }
+
+    @Override
+    public Optional<Boolean> allowNodeBefore(NodeResource nodeResourceBefore, Filter filter)
+    {
+        log.atDebug().log("Applying aspect filters on previous version of repo node id: {}", nodeResourceBefore.getId());
+        return Optional.of(allowNode(nodeResourceBefore, filter));
```
tpage-alfresco, line 61: I didn't understand the need for the optional, or even for this method. Why not just use `allowNode` for `resource` and `beforeResource`?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/RepoEventFilterHandler.java
```diff
@@ -42,13 +50,65 @@
...
+
+    /**
+     * Method handles event/node filtering. When node is denied by any of filters, a custom property "DENY_NODE: true" in Camel Exchange is created (used to filter out in main route). In case of update event: - when current version of the node is allowed but previous version was denied, then the type of the event is altered to 'Created'. - when current version of the node is denied but previous version was allowed, then the type of the event is altered to 'Deleted'.
+     *
+     * @param exchange
+     *            Camel Exchange Object
+     * @param filter
+     *            Filter configuration
+     */
+    public void handle(Exchange exchange, Filter filter)
```
krdabrowski, line 64: IMO this method should return `boolean` and `allowedNode` value instead of `void` 🤔

krdabrowski: Even class name could be then simplified to `RepoEventFilter`

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/RepoEventFilterHandler.java
```diff
@@ -42,13 +50,65 @@
...
+    {
+        final FilteringResults filteringResults = calculateFilteringResults(exchange, filter);
+        final boolean allowNode = filteringResults.allowed;
+        if (!allowNode)
+        {
+            exchange.setProperty(DENY_NODE, true);
+        }
+        filteringResults.newEventType.ifPresent(type -> exchange.getIn().setBody(camelEventMapper.alterRepoEvent(exchange, type)));
```
krdabrowski, line 72: WDYT about moving this step to `EventProcessor#proces()`?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/mapper/CamelEventMapper.java
```diff
@@ -58,4 +65,29 @@ public RepoEvent<DataAttributes<NodeResource>> repoEventFrom(Exchange exchange)
...
+    /**
+     * This method alters the original type of repo event.
+     *
+     * @param exchange
+     *            Camel exchange object
+     * @param newEventType
+     *            event type to be altered
+     * @return altered repo event
+     */
+    public RepoEvent<DataAttributes<NodeResource>> alterRepoEvent(Exchange exchange, String newEventType)
```
krdabrowski, line 78: This method could take `RepoEvent` as param instead of whole `Exchange`, WDYT?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/AspectFilterApplier.java
mpichura: This method comes from an interface which is implemented in 3 classes. 
ACS events are inconsistent in terms of passing data in current resource and resource before. In case of aspects - they are always present and "normal" filtering can be done. In case of primary hierarchy and type - they are null in `resourceBefore` but not null in `resource` if not changed.
I wanted to avoid returning `null`s - that is why the result in wrapped in `Optional`. Plus, it is easier to handle in the filtering logic.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/LiveIngesterEventHandler.java
mpichura: The case I particularly didn't like with previous approach was mixing processing with filtering.
There shouldn't be more responsibilities for filter which it had in the previous version - and now it is just a simple filter.
Processors, however, should manipulate `Exchange` and the processor I added does that under some conditions (adds a property to deny event or alters event type). 
Ideally, that processor could have been split into 2: one to determine filtering results and second one to alter the event type but I thought this solution is good enough.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/RepoEventFilterHandler.java
mpichura: See my reply on Tom's comment - long story short - I don't think we should mix filtering with processing.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/mapper/CamelEventMapper.java
mpichura: It could, however, to me, it is neater to use `Exchange` object.
Look at the usage of this method - there would be unnecessary boilerplate around it to get the repo event from `Exchange` and set it back as body.
Actually, when you raised this comment - I thought this method could be void which would handle all needed changes inside `Exchange` object so that it calling it would be simpler. However, this might not fit to `CamelEventMapper` class responsibilities (current method is already not quite fitting).
WDYT?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/RepoEventFilterHandler.java
mpichura: Actually, my previous PR has a similar part there.
We could possibly move all the filter related logic into that processor (as this single part would be quite hard to move - filtering results are `record`).
The question is whether we want to handle all of that in a single class (of course it would call/reuse classes with filtering logic)?
IMO, we should rather process in chunks, especially when we have `Camel` on board - but I generally don't think we are utilizing `Camel` well in the project and have spaghetti architecture because of that (another IMO - we may want to think about getting rid of it).
Anyway, I don't think it will be hard to move filtering logic to `EventProcessor` and reuse existing code if needed but for now I would leave that as is.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/AspectFilterApplier.java
tpage-alfresco: Ah - I didn't spot that this is different in the hierarchy filter.

How about changing the second method to take the current state as an argument:
```
boolean allowNodeBefore(boolean currentlyAllowed, NodeResource nodeResourceBefore, Filter filter)
{
    final List<String> primaryHierarchy = nodeResourceBefore.getPrimaryHierarchy();
    if (primaryHierarchy == null)
    {
        return currentlyAllowed;
    }
    return allowNode(nodeResourceBefore, filter);
}
```

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/LiveIngesterEventHandler.java
tpage-alfresco: The thing I don't like about this approach is that we're using global variables. There's no easy way to find where that property is being set. I agree that it's not nice to hide processing inside a filter though.

krdabrowski: TBH I would like to see a mixed (this x previous) approach:
`.setBody(camelEventMapper::repoEventFrom)
.filter(exchange -> repoEventFilterHandler.handle(exchange, integrationProperties.alfresco().filter())) //sets NEW_EVENT_TYPE_PROP
.process(exchange -> SecurityContextHolder.setContext(securityContext))
.process(eventProcessor::process) //processes the NEW_EVENT_TYPE_PROP and alters the RepoEvent` 
but I understand your arguments so I'm not forcing it - I'm happy when you're happy

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/RepoEventFilterHandler.java
krdabrowski: I would leave it here, I was rather thinking about setting a property like in previous PR: `exchange.setProperty(UPDATED_EVENT_TYPE_PROP, NODE_CREATED.getType());` and do the altering in 'EventProcessor#proces()'

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/LiveIngesterEventHandler.java
mpichura: `Exchange` object (message) is not global - it is request specific (unfortunately in our case rather route specific as we don't pass it further but create another one). And I would treat property of `Exchange` as metadata of message which is one of basic EIP concepts. 
I think we could take out these to lines 
```
.process(exchange -> repoEventFilterHandler.handle(exchange, integrationProperties.alfresco().filter()))
.filter(exchange -> BooleanUtils.isNotTrue(exchange.getProperty(DENY_NODE, Boolean.class)))
```
and include them (their logic to be precise) in 
```
.process(eventProcessor::process)
```
That way `EventProcessor` will be responsible for everything.
There could be another way to dela with that - set `Exchange` body to some `empty` object instead of setting `Exchange` property, That would result in:
```
.filter(exchange -> !exchange.getIn().getBody().equals("empty"))
```
Any thoughts?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/mapper/CamelEventMapper.java
krdabrowski: I would leave it as is - I agree that `void` won't fit into this class

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/LiveIngesterEventHandler.java
tpage-alfresco: I like the idea of changing the whole body. That seems easier to follow than using a property on the exchange.

Having a single `process` statement to do everything also seems easier to follow, but it wouldn't be easy to see that the message is sometimes filtered out.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/filter/AspectFilterApplier.java
mpichura: Good idea. Let me change that soon.

mpichura:
Superseded by https://github.com/Alfresco/hxinsight-connector/pull/259

PR #270, opened 2024-04-03
ACS-7473: [HxI LI Integration] Should return field values only as non-empty strings

krdabrowski:
[HxI LI Integration] Ingestion event handler should return field values only as non-empty strings
https://hyland.atlassian.net/browse/ACS-7473

### DIFF
--- a/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
+++ b/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
@@ -32,12 +32,17 @@ import static org.alfresco.hxi_connector.live_ingester.adapters.messaging.hx_ins
 import static org.alfresco.hxi_connector.live_ingester.adapters.messaging.hx_insight.model.FieldType.VALUE;
 
 import java.io.IOException;
+import java.util.Collection;
+import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import org.apache.commons.lang3.ClassUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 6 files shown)


live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializerTest.java
```diff
@@ -92,43 +96,92 @@ public void shouldSerializePropertiesToSet()
...
     @Test
-    public void canCopeWithNullUsers()
+    public void shouldSerializeObjectPropertiesToSet()
     {
         UpdateNodeEvent event = new UpdateNodeEvent(NODE_ID, CREATE)
-                .addMetadataInstruction(new NodeProperty<>(CREATED_BY_PROPERTY, null))
-                .addMetadataInstruction(new NodeProperty<>(MODIFIED_BY_PROPERTY, null));
+                .addMetadataInstruction(new NodeProperty<>(CREATED_AT_PROPERTY, 10000L))
+                .addMetadataInstruction(new NodeProperty<>("someProperty", new Object() {
+                    @Getter
```
mpichura, line 150: Nice trick with that `@Getter`

mpichura (approved):
LGTM - I had one idea regarding simplifying filtering properties traversal.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
```diff
@@ -96,6 +107,35 @@ public void serialize(UpdateNodeEvent event, JsonGenerator jgen, SerializerProvi
         }
     }
 
+    private NodeProperty<?> filterNonEmptyValueAndMap(NodeProperty<?> property)
```
mpichura, line 110: Wouldn't it be neater if you returned `Optional`? That way you could get rid of those `null`s getting returned inside conditionals.

krdabrowski: Good hint! Fixed

krdabrowski: I went event further and finally even got rid of the `Optional`

PR #280, opened 2024-04-11
ACS-6390: [HxI PH] Update Repo with predictions

krdabrowski:
https://hyland.atlassian.net/browse/ACS-6390

### DIFF
--- a/common/src/main/java/org/alfresco/hxi_connector/common/model/prediction/Prediction.java
+++ b/common/src/main/java/org/alfresco/hxi_connector/common/model/prediction/Prediction.java
@@ -1,4 +1,4 @@
-/*
+/*-
  * #%L
  * Alfresco HX Insight Connector
  * %%
@@ -23,20 +23,10 @@
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  * #L%
  */
-package org.alfresco.hxi_connector.prediction_applier;
+package org.alfresco.hxi_connector.common.model.prediction;
 
-import org.springframework.boot.SpringApplication;
-import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.retry.annotation.EnableRetry;
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
 
-@SpringBootApplication
-@EnableRetry
-@SuppressWarnings("PMD.UseUtilityClass")
-public class PredictionApplierApplication
-{
-
-    public static void main(String[] args)
-    {
-        SpringApplication.run(PredictionApplierApplication.class, args);
-    }
-}
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+public record Prediction(String id, String objectId)
+{}
--- a/common/src/main/java/org/alfresco/hxi_connector/common/model/repository/Node.java
+++ b/common/src/main/java/org/alfresco/hxi_connector/common/model/repository/Node.java
@@ -1,4 +1,4 @@
-/*
+/*-
  * #%L
  * Alfresco HX Insight Connector
  * %%
@@ -23,20 +23,19 @@
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  * #L%
  */
-package org.alfresco.hxi_connector.prediction_applier;
+package org.alfresco.hxi_connector.common.model.repository;
 
-import org.springframework.boot.SpringApplication;
-import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.retry.annotation.EnableRetry;
+import java.util.List;
+import jakarta.validation.constraints.NotBlank;
 
-@SpringBootApplication
-@EnableRetry
-@SuppressWarnings("PMD.UseUtilityClass")
-public class PredictionApplierApplication
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record Node(@NotBlank String id, @JsonProperty("aspectNames") List<String> aspects)
 {
-
-    public static void main(String[] args)
+    public Node(String id)
     {
-        SpringApplication.run(PredictionApplierApplication.class, args);
+        this(id, null);
     }
 }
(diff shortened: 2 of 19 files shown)


common/src/main/java/org/alfresco/hxi_connector/common/model/repository/Node.java
```diff
@@ -0,0 +1,41 @@
...
+package org.alfresco.hxi_connector.common.model.repository;
+
+import java.util.List;
+import jakarta.validation.constraints.NotBlank;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record Node(@NotBlank String id, @JsonProperty("aspectNames") List<String> aspects)
```
tpage-alfresco, line 35: I expected this information to be inside an object called `Prediction` or maybe `NodePrediction`.

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/hxinsight/PredictionMapper.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.alfresco.hxi_connector.common.model.prediction.Prediction;
+import org.alfresco.hxi_connector.common.model.repository.Node;
+
+@Component
+@RequiredArgsConstructor
+public class PredictionMapper
+{
+
+    public Node map(Prediction prediction)
```
tpage-alfresco, line 41: I thought there would be three conversion steps - something like:

1. HxI prediction batch deserialisation object (from REST API) -> Internal prediction batch object
2. Internal prediction batch object -> Internal node prediction task object (gets put onto AMQ)
3. Internal node prediction task object (retrieved from AMQ) -> Alfresco node update request (sent to REST API)

I'm not sure which of the two this step is, or whether this is actually a stage in the middle of step 3?

common/src/main/java/org/alfresco/hxi_connector/common/model/prediction/Prediction.java
```diff
@@ -0,0 +1,32 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.hxi_connector.common.model.prediction;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+public record Prediction(String id, String objectId)
```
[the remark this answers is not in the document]
krdabrowski: Exactly - this is only a start point - I've focused more on passing some value, not the value itself, as this wasn't (or still isn't 🤔) clear

common/src/main/java/org/alfresco/hxi_connector/common/model/repository/Node.java
krdabrowski: This should represent the object which is passed to `/nodes` API - if you think this is not clear enough I don't mind to rename it - `NodePrediction` should be OK IMO

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/hxinsight/PredictionMapper.java
krdabrowski: This should correspond to 3. stage - this is mapping of an object received from AMQ and then passed to repo's `/nodes` API. I didn't want to introduce this mapping at all - but I came up with an idea of adding also #281 so I needed some MVP of this prediction flow, and for that I needed some really basic mapping

mpichura (commented):
I didn't have enough time to go through all the changes but left some comments anyway.

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/hxinsight/PredictionListener.java
```diff
@@ -25,32 +25,49 @@
...
 
     @Override
     public void configure()
     {
         from(endpoint)
                 .routeId(ROUTE_ID)
                 .log(LoggingLevel.DEBUG, log, "Prediction body: ${body}")
+                .unmarshal()
+                .json(JsonLibrary.Jackson, Prediction.class)
+                .process(this::processPrediction)
```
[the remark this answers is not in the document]
krdabrowski: This is a good hint - tomorrow I will try to play with it

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/hxinsight/PredictionListener.java
```diff
@@ -25,32 +25,49 @@
...
+    private final NodesClient nodesClient;
 
     @Override
     public void configure()
     {
         from(endpoint)
                 .routeId(ROUTE_ID)
                 .log(LoggingLevel.DEBUG, log, "Prediction body: ${body}")
+                .unmarshal()
+                .json(JsonLibrary.Jackson, Prediction.class)
```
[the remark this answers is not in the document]
krdabrowski: This is true, but I think this should be done in a separate PR - this task is about calling Repo API

common/src/main/java/org/alfresco/hxi_connector/common/model/repository/Node.java
mpichura: I think `Node` is okay here as it is the node that gets send to ACS- but that's just my opinion.

tpage-alfresco: I didn't realise this was the object that was being passed to the `/nodes` API. Please leave it as is. I think it will become more obvious as we add support for populating more fields.

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/hxinsight/PredictionListener.java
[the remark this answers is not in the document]
tpage-alfresco: Polling should be done in https://hyland.atlassian.net/browse/ACS-7544

I'll add your suggestions there Maciej :+1:

tpage-alfresco (approved):
LGTM

mpichura (approved):
LGTM

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/hxinsight/PredictionListener.java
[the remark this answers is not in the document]
krdabrowski: I've created second PR with more Camel-way implementation: https://github.com/Alfresco/hxinsight-connector/pull/287, let me know WDYT about it

PR #282, opened 2024-04-11
ACS-7525: test with different repo versions

Pawel-608:


krdabrowski (approved):
Nice job!

mpichura (approved):
LGTM.
Can we move the test which verifies if our JAMP is properly installed to E2E-tests module? Or at least test that part too?

Pawel-608:
I asked Krystian about this and he told me that he would prefer to keep it in extension module, but he also told me that he's preparing e2e test that will test even more than this test so imo it is not necessary. We can also discuss it later if you want

PR #277, opened 2024-04-11
ACS-6906: Add mime type configuration for pdf transforms.

mpichura:
Adding empty configuration for mime type mappings. 
However, this configuration should allow to set up the mappings in following way:
``` yml
    mime-type:
      mapping:
        '[image/png]': "image/png"
        '[image/bmp]': "image/png"
        '[image/gif]': "image/png"
        '[image/raw]': "image/png"
        '[image/*]': "image/jpeg"
        '[*]': "application/pdf"
```
In the next PRs I will utilize this configuration to request transforms to mime types other than PDF (with support of exact matching, partial wildcards for subtypes and wildcards).

PR #284, opened 2024-04-11
ACS-6905: Handling various mime type configurations for transforms.

mpichura:
A follow up to https://github.com/Alfresco/hxinsight-connector/pull/277
If no mime-type mapping is configured, then every content will be mapped to `application/pdf`.
If configured, mapping is done basing on following steps:
- look up exact mime-type mapping in configuration
- if above not found - look up wildcard subtype (`type/*`, e.g., `image/*`, `application/*` etc.) mapping in configuration
- if above not found - look up full wildcard mapping in configuration (`*`)

For instance, given following mapping configuration:
``` yml
    mime-type:
      mapping:
        '[image/png]': "image/png"
        '[image/bmp]': "image/png"
        '[image/gif]': "image/png"
        '[image/raw]': "image/png"
        '[image/*]': "image/jpeg"
        '[*]': "application/pdf"
```
` image/png`, `image/bmp`, `image/gif`, `image/raw` content type will request transforms to `image/png` content type.
All other `image` content types will be transformed to `image/jpeg` content type.
All other content types will be transformed to `application/pdf` content type.

### DIFF
--- a/live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/bulk_ingester/BulkIngesterEventNonMatchingContentMappingIntegrationTest.java
+++ b/live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/bulk_ingester/BulkIngesterEventNonMatchingContentMappingIntegrationTest.java
@@ -1,0 +1,108 @@
+/*
+ * #%L
+ * Alfresco HX Insight Connector
+ * %%
+ * Copyright (C) 2023 - 2024 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.hxi_connector.live_ingester.domain.usecase.e2e.bulk_ingester;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.springframework.boot.test.context.SpringBootTest;
+
+import org.alfresco.hxi_connector.live_ingester.util.E2ETestBase;
+
+@SpringBootTest(properties = {"alfresco.transform.mime-type.mapping.[image/png]=image/png",
+        "alfresco.transform.mime-type.mapping.[image/bmp]=image/png",
+        "alfresco.transform.mime-type.mapping.[image/raw]=image/png",
+        "alfresco.transform.mime-type.mapping.[image/gif]=image/png",
+        "alfresco.transform.mime-type.mapping.[image/*]=image/jpeg",
+        "alfresco.transform.mime-type.mapping.[application/*]=application/pdf",
+        "logging.level.org.alfresco=DEBUG"})
+@SuppressWarnings("PMD.JUnitTestsShouldIncludeAssert")
+public class BulkIngesterEventNonMatchingContentMappingIntegrationTest extends E2ETestBase
+{
+
+    @ParameterizedTest
+    @ValueSource(strings = {"text/plain", "text/html", "text/richtext"})
+    void givenExactAndWildcardMimeTypeMappingForContentConfigured_whenContentWithNotMatchingTypeIngested_thenProcessWithoutTransformRequest(
+            String sourceMimeType)
+    {
+        // given
+        containerSupport.prepareHxInsightToReturnSuccess();
+
+        // when
+        String repoEvent = """
+                {
+                  "nodeId": "37be157c-741c-4e51-b781-20d36e4e335a",
+                  "contentInfo": {
+                    "contentSize": 330,
+                    "encoding": "ISO-8859-1",
+                    "mimetype": "%s"
+                  },
+                  "properties": {
+                    "cm:name": "dashboard.xml",
+                    "cm:isContentIndexed": true,
+                    "cm:isIndexed": false,
+                    "createdAt": 1308061016,
+                    "type": "cm:content",
+                    "createdBy": "admin",
+                    "modifiedBy": "hr_user",
+                    "aspectsNames": [
+                      "cm:indexControl",
+                      "cm:auditable"
+                    ]
+                  }
+                }""".formatted(sourceMimeType);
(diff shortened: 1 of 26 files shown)


live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/bulk_ingester/BulkIngesterEventNonMatchingContentMappingIntegrationTest.java
```diff
@@ -0,0 +1,108 @@
...
+                            "size": 330,
+                            "content-type": "%s"
+                          }
+                        }
+                      }
+                    }
+                  }
+                ]""".formatted(sourceMimeType);
+        containerSupport.expectHxIngestMessageReceived(expectedBody);
+        containerSupport.verifyNoATSRequestReceived(200);
```
tpage-alfresco, line 106: Nice extra feature :+1: We can avoid sending everything to ATS.

live-ingester/src/main/resources/application.yml
```diff
@@ -34,9 +34,6 @@ alfresco:
     shared-file-store:
       host: http://localhost
       port: 8099
-    mime-type:
-      mapping:
-        '[*]': "application/pdf"
```
tpage-alfresco, line 39: I still think it's worth providing some default settings here. Maybe something similar to what you've used in the e2es above? Possibly exactly what you've done for images and then a curated list of mime-types that will convert to pdf?

tpage-alfresco: Or alternatively we could make the default no-config behaviour more complex, but I think just including what we think customers will want here is best.

tpage-alfresco (approved):
Looks really good to me. I think we could provide more value by having a better default for the config file, but happy to approve as is.

live-ingester/src/main/resources/application.yml
mpichura: Honestly, I had several ideas on how to tackle that. 
I ended up with being sort of backwards compatible - when there is no config in the properties - the target mime-type is always `application/pdf` (we used to do that until this change came in).
Any other configs reflected in properties can break that "compatibility" and I thought they'd have to be consciously set up.
I am happy to put E2Es alike config here
``` yml
    mime-type:
      mapping:
        '[image/png]': "image/png"
        '[image/bmp]': "image/png"
        '[image/gif]': "image/png"
        '[image/raw]': "image/png"
        '[image/*]': "image/jpeg"
        '[*]': "application/pdf"
```
Shall I do that?

tpage-alfresco: I think that's a reasonable first version. I did wonder if we could try to find a list of all mimetypes that can be converted to pdf by the default ATS set up, and then leave everything else so that it won't attempt conversion. I'm happy with this suggestion though.

mpichura: I've found [this](https://docs.alfresco.com/content-services/latest/admin/transformations/) - quite a long list of possible transformations. I'll have a look at that and figure out whether it can be compressed to a reasonable list.

tpage-alfresco: Nice find. Looks like we could get a reasonably improvement by defaulting to all of `[application/*]` and `[text/*]` converted to pdf.

mpichura: Yep, will do that - seems to be a reasonable default.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/properties/Transform.java
```diff
@@ -27,16 +27,28 @@
...
+public record Transform(@NotNull Request request, @NotNull Response response, @NotNull SharedFileStore sharedFileStore, MimeType mimeType)
 {
+
+    @ConstructorBinding
+    public Transform(@NotNull Request request, @NotNull Response response, @NotNull SharedFileStore sharedFileStore, MimeType mimeType)
+    {
+        this.request = request;
+        this.response = response;
+        this.sharedFileStore = sharedFileStore;
+        this.mimeType = mimeType != null ? mimeType : new MimeType(Collections.emptyMap());
```
krdabrowski, line 49: I wonder if this constructor could be replaced with:
```
public Transform
{
    mimeType = requireNonNullElse(mimeType, new MimeType(Collections.emptyMap()));
}
```
🤔

krdabrowski (approved):
Great job!

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/mapper/MimeTypeMapper.java
```diff
@@ -0,0 +1,77 @@
...
+public class MimeTypeMapper
+{
+
+    public static final String EMPTY_MIME_TYPE = "";
+    static final String DEFAULT_MIME_TYPE = "application/pdf";
+    private final IntegrationProperties integrationProperties;
+
+    public String mapMimeType(String inputType)
+    {
+        Map<String, String> mappings = MapUtils.emptyIfNull(integrationProperties.alfresco().transform().mimeType().mapping());
```
krdabrowski, line 48: `.mimeType().mapping()` shouldn't be with 's' at the end (in plural form)?

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/properties/Transform.java
mpichura: Thanks for the hint - it works.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/mapper/MimeTypeMapper.java
mpichura: I honestly don't know how to tackle that as there seem not to be solid conventions for properties keys naming (at least that I know of). 
So the question here is whether to configure the key with singular (as it is now):
``` yml
    mime-type:
      mapping:
        '[image/png]': "image/png"
        ....
```
or with plural:
``` yml
    mime-type:
      mappings:
        '[image/png]': "image/png"
        ....
```
I am okay with any. @tpage-alfresco @Pawel-608 WDYT?

krdabrowski: IMO plural indicates, that there can be more then one value (or pair), similarly to List or Set

tpage-alfresco: Mapping or mappings are both fine from an English pov. A "mapping" could be a single source-target pair, or it could be the entire instruction set.

For the filter you could change it to `allows` and `denies`, but IMO it reads as a verb rather than a plural.

krdabrowski: Maybe `aspects`, `types` and `paths` should be in plural instead of `allows` and `denys` 🤔

tpage-alfresco: I think this is fine tbh. Customers are going to need to consult the docs anyway, so I don't think there's a huge value in changing this.

tpage-alfresco:
I like this solution for default MIME types :+1:

krdabrowski (approved):
Still LGTM

PR #287, opened 2024-04-15
ACS-6390: [HxI PH] Update Repo with predictions - Camel way

krdabrowski:
https://hyland.atlassian.net/browse/ACS-6390

Camel approach replacement for PR https://github.com/Alfresco/hxinsight-connector/pull/280

### DIFF
--- a/prediction-applier/src/test/java/org/alfresco/hxi_connector/prediction_applier/repository/NodesClientTest.java
+++ b/prediction-applier/src/test/java/org/alfresco/hxi_connector/prediction_applier/repository/NodesClientTest.java
@@ -30,6 +30,7 @@ import static org.apache.camel.builder.AdviceWith.adviceWith;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.catchThrowable;
 
+import static org.alfresco.hxi_connector.prediction_applier.repository.NodesClient.NODES_DIRECT_ENDPOINT;
 import static org.alfresco.hxi_connector.prediction_applier.repository.NodesClient.ROUTE_ID;
 
 import java.util.Set;
@@ -96,12 +97,12 @@ class NodesClientTest
         // given
         String nodeId = "node-id";
         String aspect = "aspect-name";
-        Node node = new Node(nodeId, null);
+        Node node = new Node(nodeId, Set.of(aspect));
         mockEndpointWillRespondWith(200, createResponseBodyWith(nodeId, aspect));
         mockEndpoint.expectedMessageCount(1);
 
         // when
-        Node actualNode = producerTemplate.to(NodesClient.NODES_DIRECT_ENDPOINT)
+        Node actualNode = producerTemplate.to(NODES_DIRECT_ENDPOINT)
                 .withBody(node)
                 .request(NodeEntry.class)
                 .node();
@@ -119,7 +120,7 @@ class NodesClientTest
         mockEndpoint.expectedMessageCount(1);
 
         // when
-        Throwable thrown = catchThrowable(() -> producerTemplate.to(NodesClient.NODES_DIRECT_ENDPOINT).request());
+        Throwable thrown = catchThrowable(() -> producerTemplate.to(NODES_DIRECT_ENDPOINT).request());
 
         // then
         mockEndpoint.assertIsSatisfied();
@@ -136,7 +137,7 @@ class NodesClientTest
         mockEndpoint.expectedMessageCount(RETRY_ATTEMPTS + 1);
 
         // when
-        Throwable thrown = catchThrowable(() -> producerTemplate.to(NodesClient.NODES_DIRECT_ENDPOINT).request());
+        Throwable thrown = catchThrowable(() -> producerTemplate.to(NODES_DIRECT_ENDPOINT).request());
 
         // then
         mockEndpoint.assertIsSatisfied();


common/src/main/java/org/alfresco/hxi_connector/common/exception/EndpointClientErrorException.java
```diff
@@ -0,0 +1,44 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.hxi_connector.common.exception;
+
+public class EndpointClientErrorException extends RuntimeException
```
tpage-alfresco, line 28: I'm making the exact same change in my PR. Let's extract the exception movement into a separate PR.

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/repository/NodesClient.java
```diff
@@ -0,0 +1,123 @@
...
+import org.alfresco.hxi_connector.common.exception.EndpointServerErrorException;
+import org.alfresco.hxi_connector.common.model.repository.NodeEntry;
+import org.alfresco.hxi_connector.common.util.ErrorUtils;
+import org.alfresco.hxi_connector.prediction_applier.config.NodesApiProperties;
+
+@Component
+@RequiredArgsConstructor
+public class NodesClient extends RouteBuilder
+{
+    public static final String DIRECT_ENDPOINT = "direct:" + NodesClient.class.getSimpleName();
```
mpichura, line 54: I think constant name could be more explanatory.
```suggestion
    public static final String NODES_ENDPOINT = "direct:" + NodesClient.class.getSimpleName();
```
or
```suggestion
    public static final String NODES_DIRECT_ENDPOINT = "direct:" + NodesClient.class.getSimpleName();
```

prediction-applier/src/test/java/org/alfresco/hxi_connector/prediction_applier/hxinsight/PredictionListenerTest.java
```diff
@@ -0,0 +1,121 @@
...
+    FluentProducerTemplate producerTemplate;
+
+    PredictionMapper predictionMapperMock;
+
+    @BeforeAll
+    @SneakyThrows
+    void beforeAll()
+    {
+        camelContext = new DefaultCamelContext();
+        predictionMapperMock = mock(PredictionMapper.class);
```
mpichura, line 74: Just an idea - not crucial. Since this test is a bit "bigger" than a unit test - maybe `predictionMapperMock` could be a new instance of `PredictionMapper`?

prediction-applier/src/test/java/org/alfresco/hxi_connector/prediction_applier/hxinsight/PredictionListenerTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.alfresco.hxi_connector.common.model.prediction.Prediction;
+import org.alfresco.hxi_connector.common.model.repository.Node;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class PredictionListenerTest
+{
+    private static final String TEST_ENDPOINT = "direct:predictions-queue";
+    private static final String MOCK_ENDPOINT = "mock:direct-nodes-endpoint";
+
+    CamelContext camelContext;
```
mpichura, line 63: Any particular reason this field and 3 following ones are package-private (could be private I guess)?

prediction-applier/src/test/java/org/alfresco/hxi_connector/prediction_applier/repository/NodesClientTest.java
```diff
@@ -0,0 +1,172 @@
...
+import org.alfresco.hxi_connector.common.model.repository.NodeEntry;
+import org.alfresco.hxi_connector.prediction_applier.config.NodesApiProperties;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class NodesClientTest
+{
+    private static final String MOCK_ENDPOINT = "mock:repo-nodes-endpoint";
+    private static final int RETRY_ATTEMPTS = 3;
+
+    CamelContext camelContext;
```
mpichura, line 64: Any particular reason this field and 2 following ones are package-private (could be private I guess)?

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/repository/NodesClient.java
```diff
@@ -0,0 +1,123 @@
...
+            .stop();
+
+        from(DIRECT_ENDPOINT)
+            .setHeader(NODE_ID_HEADER, simple("${body.id}"))
+            .marshal()
+            .json(JsonLibrary.Jackson)
+            .to(RETRYABLE_ROUTE)
+            .end();
+
+        from(RETRYABLE_ROUTE)
```
mpichura, line 97: Is this route separated out (from the one starting on `DIRECT_ENDPOINT`) because of retry handling?

mpichura (approved):
Generally, LGTM - my remarks are minor or are just ideas.
I wonder if the tests utilizing `Camel` are visibly faster (or slower)?
I also wonder what are your thoughts on going more `Camel` way...

common/src/main/java/org/alfresco/hxi_connector/common/exception/EndpointClientErrorException.java
krdabrowski: PR ready for review: https://github.com/Alfresco/hxinsight-connector/pull/290 😄

prediction-applier/src/test/java/org/alfresco/hxi_connector/prediction_applier/hxinsight/PredictionListenerTest.java
krdabrowski: This is an option, but if `PredictionMapper` will become dependent on some else, this will require also change in this test and I wanted to avoid such a situation

prediction-applier/src/test/java/org/alfresco/hxi_connector/prediction_applier/hxinsight/PredictionListenerTest.java
krdabrowski: This is something that @tpage-alfresco taught me - from the test perspective components access level is not important and this, at least in theory should make the test a bit easier to analyze and understand

krdabrowski: In the other hand I must admit, that at the beginning this way of writing seemed strange to me, but now I'm already used to it

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/repository/NodesClient.java
krdabrowski: Yes - Camel by default retries only the point where exception was thrown, not the whole route. Because in our case exception is thrown after call to endpoint:
```
.toD(URI_PATTERN.formatted(nodesApiProperties.baseUrl(), NODE_ID_HEADER, nodesApiProperties.username(), nodesApiProperties.password()))
.choice()
.when(header(HTTP_RESPONSE_CODE).isNotEqualTo(String.valueOf(EXPECTED_STATUS_CODE)))
    .process(this::throwExceptionOnUnexpectedStatusCode)
```
I had to extract this to separate sub-route and use the `.errorHandler(noErrorHandler())`

krdabrowski:
Tests are much faster - right now using `CamelContxt`, `MockEndpoint`s and `TestInstance.Lifecycle.PER_CLASS` they run in about 300 ms comparing to ~1.2 s with WireMock. Disadvantage of this approach is how calls to `MockEndpoint` are verified - quite strange IMO.
WDYM with going even more `Camel` way?

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/repository/NodesClient.java
```diff
@@ -0,0 +1,123 @@
...
+
+@Component
+@RequiredArgsConstructor
+public class NodesClient extends RouteBuilder
+{
+    public static final String NODES_DIRECT_ENDPOINT = "direct:" + NodesClient.class.getSimpleName();
+    private static final String RETRYABLE_ROUTE = "direct:retryable-" + NodesClient.class.getSimpleName();
+    static final String ROUTE_ID = "repository-nodes";
+    private static final String NODE_ID_HEADER = "nodeId";
+    private static final String URI_PATTERN = "%s/alfresco/api/-default-/public/alfresco/versions/1/nodes/${headers.%s}?httpMethod=PUT&authMethod=Basic&authUsername=%s&authPassword=%s&authenticationPreemptive=true&throwExceptionOnFailure=false";
```
tpage-alfresco, line 58: This is fine for now, but we're going to need another ticket to handle other authentication types.

prediction-applier/src/test/java/org/alfresco/hxi_connector/prediction_applier/hxinsight/PredictionListenerTest.java
tpage-alfresco: I'm still not consistent in this. In theory this makes sense to me, but it seems harder to remember _not_ to write private!

prediction-applier/src/test/java/org/alfresco/hxi_connector/prediction_applier/repository/NodesClientTest.java
```diff
@@ -0,0 +1,172 @@
...
+        camelContext.stop();
+    }
+
+    @Test
+    void testUpdateNode() throws InterruptedException
+    {
+        // given
+        String nodeId = "node-id";
+        String aspect = "aspect-name";
+        Node node = new Node(nodeId, null);
```
tpage-alfresco, line 99: I think you should provide `aspect` in this constructor, since it's meant to have been predicted by HxI?

prediction-applier/src/test/java/org/alfresco/hxi_connector/prediction_applier/hxinsight/PredictionListenerTest.java
krdabrowski: In the other hand it's a good habit write the `private` keyword

tpage-alfresco (approved):
LGTM

prediction-applier/src/test/java/org/alfresco/hxi_connector/prediction_applier/repository/NodesClientTest.java
krdabrowski: This is a node meant to be passes do `/nodes` API, but still good catch - I will fix it to reflect more reality

PR #292, opened 2024-04-16
ACS-7544: Rename hxinsight package to hxi_connector

Pawel-608:


prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/hxi_connector/PredictionListener.java
```diff
@@ -23,7 +23,7 @@
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  * #L%
  */
-package org.alfresco.hxi_connector.prediction_applier.hxinsight;
+package org.alfresco.hxi_connector.prediction_applier.hxi_connector;
```
krdabrowski, line 26: Is it required to have string "hxi_connector" twice package name?

PR #295, opened 2024-04-17
ACS-7543 Create common module for Hx Authentication.

tpage-alfresco:


tpage-alfresco:
We previously discussed whether to create a module for this or not and decided it wasn't worth the effort. However we since found out that we're likely to also need common authentication for ACS (and it may overlap with Hx authentication, since both can use OAuth2). Consequently I created a new module for the common code.

krdabrowski (approved):
LGTM - nice job!

mpichura (approved):
Nice work

PR #300, opened 2024-04-18
ACS-7450 Fix default MIME type mapping.

tpage-alfresco:
We now omit the mappings by default, which causes the default mappings to be used in the `MimeTypeMapper`.


azakrzewski-hy (approved):
LGTM

PR #303, opened 2024-04-19
Split execution of different subprojects test

Pawel-608:


.github/workflows/ci.yml
```diff
@@ -89,7 +93,7 @@ jobs:
           password: ${{ secrets.QUAY_PASSWORD }}
       - uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v1.35.2
       - name: "Test application"
-        run: mvn ${{ env.MAVEN_CLI_OPTS }} clean verify -pl '!e2e-test'
+        run: mvn ${{ env.MAVEN_CLI_OPTS }} clean verify -pl '${{ matrix.subproject }}'
```
krdabrowski, line 96: Down side of this approach is: when someone will make changes in several modules, which depends in each other the test will fail, because only one module will be build within job, and changes from other modules won't be seen. This could be resolved by adding the `-am` flag, but this may result in building more then one module, depending on the dependencies list

.github/workflows/ci.yml
```diff
@@ -78,6 +78,10 @@ jobs:
     if: >
       !contains(github.event.head_commit.message, '[skip tests]') &&
       !contains(github.event.head_commit.message, '[skip build]')
+    strategy:
+      matrix:
+        subproject: [ "bulk-ingester", "common", "common-authentication", "common-test",
+                      "live-ingester", "prediction-applier", "prediction-applier-extension" ]
```
krdabrowski, line 84: `distribution` module is missing

.github/workflows/ci.yml
Pawel-608: Good point! I'll add the flag and check if it will make jobs execution much slower...

.github/workflows/ci.yml
Pawel-608: there is no tests there, but maybe I should add it 🤔

.github/workflows/ci.yml
mpichura: I think adding `-am` option is a good hint here, also `-DfailIfNoTests=false` could be useful if module has no tests at all (at least AFAIR)

.github/workflows/ci.yml
krdabrowski: Module build phase is also kind of a test if there are not any compilation/packaging errors

.github/workflows/ci.yml
Pawel-608: verify on distribution doesn't work for some reason so I extracted it to separate job

krdabrowski (approved):
LGTM

PR #307, opened 2024-04-19
Improve live-ingester integration tests

Pawel-608:


live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/bulk_ingester/BulkIngesterEventNonMatchingContentMappingIntegrationTest.java
```diff
@@ -23,29 +23,25 @@
...
+import org.springframework.test.context.TestPropertySource;
 
-import org.alfresco.hxi_connector.live_ingester.util.E2ETestBase;
+import org.alfresco.hxi_connector.live_ingester.util.camel.CamelTestBase;
 
-@SpringBootTest(properties = {"logging.level.org.alfresco=DEBUG"})
+@TestPropertySource(properties = {"alfresco.transform.mime-type.mapping.[text/*]="})
 @SuppressWarnings("PMD.JUnitTestsShouldIncludeAssert")
-public class BulkIngesterEventNonMatchingContentMappingIntegrationTest extends E2ETestBase
+public class BulkIngesterEventNonMatchingContentMappingIntegrationTest extends CamelTestBase
```
tpage-alfresco, line 36: This makes me nervous. We're moving from a known quantity - testing the application E2E by using real external interfaces, to testing internally by presumably using hooks into Camel. I would definitely prefer to stick to testing with external interfaces and internally with unit tests.

Pawel-608: meh, these aren't real E2E tests, these are idk "component tests"? "behavioural tests"? that have a lot of connections to outside world

way I propose to write them makes them:
- faster
- easier to maintain
- more reliable [ we do not need to use wiremock anymore]
- they give exactly same value as we have from these tests now

so imo it is better approach

@mpichura wdyt? should I close this pr or we should discuss it (if 2 people don't like the idea I'll give up😅)

PR #309, opened 2024-04-22
ACS-6826 add method to pull enterprise repository image

azakrzewski-hy:
[ACS-6826](https://hyland.atlassian.net/browse/ACS-6826)
This method is added so enterprise repository image would be downloaded when local run for test, which requires it, is performed.

[ACS-6826]: https://hyland.atlassian.net/browse/ACS-6826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mpichura (approved):
LGTM, although I think we may also try to find the reason it is needed.

PR #324, opened 2024-04-30
ACS-6391 Add data model for HxI predictions.

tpage-alfresco:
The property name that the prediction applies to will be stored in the child node's `cm:name` field. This will have the colon replaced by an underscore (as a colon is not valid in a name field) - e.g. `cm:description` will be `cm_description`. This matches the behaviour when creating field names in Elasticsearch.

Since there is a uniqueness check for `cm:name` then this ensures we have at most one (latest) prediction for each property.

### DIFF
--- a/prediction-applier-extension/src/main/resources/alfresco/module/alfresco-hxinsight-connector-prediction-applier-extension/model/hxi-model.xml
+++ b/prediction-applier-extension/src/main/resources/alfresco/module/alfresco-hxinsight-connector-prediction-applier-extension/model/hxi-model.xml
@@ -15,10 +15,96 @@
         <namespace uri="http://www.alfresco.org/model/hxi/1.0" prefix="hxi"/>
     </namespaces>
 
+    <constraints>
+        <constraint name="hxi:updateTypeList" type="LIST">
+            <parameter name="allowedValues">
+                <list>
+                    <value>AUTOFILL</value>
+                    <value>AUTOCORRECT</value>
+                </list>
+            </parameter>
+            <parameter name="caseSensitive">
+                <value>false</value>
+            </parameter>
+        </constraint>
+    </constraints>
+
+    <types>
+        <type name="hxi:prediction">
+            <title>Prediction</title>
+            <parent>cm:cmobject</parent>
+
+            <properties>
+                <property name="hxi:predictionDateTime">
+                    <title>Prediction Date Time</title>
+                    <description>Time that the prediction was generated</description>
+                    <type>d:datetime</type>
+                    <mandatory>true</mandatory>
+                </property>
+                <property name="hxi:confidenceLevel">
+                    <title>Confidence Level</title>
+                    <description>Prediction confidence from Hx Insight</description>
+                    <type>d:float</type>
+                    <mandatory>true</mandatory>
+                </property>
+                <property name="hxi:modelId">
+                    <title>Prediction Model Id</title>
+                    <description>Identified for the model used by Hx Insight</description>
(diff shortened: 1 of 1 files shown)


krdabrowski (approved):
LGTM

prediction-applier-extension/src/main/resources/alfresco/module/alfresco-hxinsight-connector-prediction-applier-extension/model/hxi-model.xml
```diff
@@ -15,10 +15,96 @@
...
+                <property name="hxi:modelId">
+                    <title>Prediction Model Id</title>
+                    <description>Identified for the model used by Hx Insight</description>
+                    <type>d:text</type>
+                    <mandatory>true</mandatory>
+                </property>
+                <property name="hxi:predictionValue">
+                    <title>Prediction Value</title>
+                    <description>Value predicted by Hx Insight</description>
+                    <type>d:any</type>
```
aborroy, line 59: Why to use ANY instead of NODE_REF?
Is the value stored for ANY properties searchable in Solr/Elasticsearch?

prediction-applier-extension/src/main/resources/alfresco/module/alfresco-hxinsight-connector-prediction-applier-extension/model/hxi-model.xml
```diff
@@ -15,10 +15,96 @@
...
+                    <mandatory>true</mandatory>
+                </property>
+                <property name="hxi:updateType">
+                    <title>Update Type</title>
+                    <description>Whether the prediction exceeded the threshold for autofill or autocorrect</description>
+                    <type>d:text</type>
+                    <mandatory>true</mandatory>
+                    <constraints>
+                        <constraint ref="hxi:updateTypeList" />
```
aborroy, line 74: Is this list (AUTOFILL, AUTOCORRECT) closed? Is it possible that additional values may be used in the future? If this is the case, updating the content model should be required when using a new version of the HXI service. That is not a major concern, just raising a dependency that may be introduced by this constraint.

aborroy (changes requested):
Minor comments.

prediction-applier-extension/src/main/resources/alfresco/module/alfresco-hxinsight-connector-prediction-applier-extension/model/hxi-model.xml
tpage-alfresco: The value here could be a boolean, number, string, etc. This is the predicted value from Hx Insight.

prediction-applier-extension/src/main/resources/alfresco/module/alfresco-hxinsight-connector-prediction-applier-extension/model/hxi-model.xml
tpage-alfresco: That's a good point. We were thinking of introducing "suggestions" - lower confidence predictions that should not replace the existing value.

tpage-alfresco: Discussed with Angel that:
1. We can update or remove this constraint in future with a model update (requires module reinstall and ACS restart)
2. We are very likely to require a module update anyway if we're adding new update types (e.g. suggestions).

Consequently it seems fine to leave this as it is.

PR #328, opened 2024-05-02
ACS-7544: [HxI PH] Poll Hx Insight endpoint and add predictions to AMQ - added DC config

krdabrowski:
https://hyland.atlassian.net/browse/ACS-7544


tpage-alfresco (approved):
LGTM

PR #347, opened 2024-05-10
WIP: ACS-7586: Make code changes once prediction API is published

krdabrowski:
https://hyland.atlassian.net/browse/ACS-7586

PR contains working PoC allowing to consume actual HxI predictions API. It supports both, batches and predictions pagination.
What is required to do/fix:
- tests wasn't touched almost at all (only adjusted to models changes)
- Camel monster route should be split into smaller parts
- `quartz` route start should be delayed to run the auth first
- logging needs to be adjusted
- notifying HxI about batch status change
- probably many more

### DIFF
--- a/prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/model/prediction/PredictionBatch.java
+++ b/prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/model/prediction/PredictionBatch.java
@@ -23,21 +23,22 @@
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  * #L%
  */
-package org.alfresco.hxi_connector.prediction_applier.hx_insight;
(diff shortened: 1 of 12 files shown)


prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/model/prediction/PredictionBatch.java
```diff
@@ -0,0 +1,44 @@
...
+ * #L%
+ */
+package org.alfresco.hxi_connector.prediction_applier.model.prediction;
+
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record PredictionBatch(
+        @JsonProperty("_id") String id,
```
krdabrowski, line 33: I'm not sure if all fields are required - right now only `id` is used, so maybe we could remove all except `id`? 🤔

PR #356, opened 2024-05-20
ACS-7708: send info to hxi about confirmed predictions

Pawel-608:


azakrzewski-hy (approved):
LGTM, one question asked

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/EventProcessor.java
```diff
@@ -73,22 +76,38 @@ public void process(Exchange exchange)
...
+        }
+
+        handleMetadataPropertiesChange(event);
+        handleContentChange(event);
+        handleNodeDeleteEvent(event);
+    }
+
+    private void handlePredictionNodeEvent(RepoEvent<DataAttributes<NodeResource>> event)
+    {
+        ensureThat(isPredictionNodeEvent(event), "Event is not a prediction node event");
```
tpage-alfresco, line 105: This is fine, but it does seem like we're using the wrong pattern. The calling code literally just made exactly the same check.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/EventUtils.java
```diff
@@ -89,29 +97,108 @@ public static EventType getEventType(RepoEvent<DataAttributes<NodeResource>> eve
...
+        // For events other than create or update then we do not need to process the content.
+        return false;
+    }
+
+    public static boolean wasNodePropertyChangedTo(RepoEvent<DataAttributes<NodeResource>> event, String propertyName, Object value)
+    {
+        Map<String, Serializable> oldProperties = event.getData().getResourceBefore().getProperties();
+        Map<String, Serializable> newProperties = event.getData().getResource().getProperties();
+
+        if (oldProperties == null || newProperties == null)
```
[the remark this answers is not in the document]
tpage-alfresco: You right Adam, but in fact the event contract says that if a property has changed in an update then we will get both values (old and new). If we're missing the old value in an update event then it means that the value stayed the same.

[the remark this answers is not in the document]
tpage-alfresco: It would be good to have some unit tests of this method though

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/EventProcessorTest.java
```diff
@@ -298,6 +302,42 @@ void shouldNotProcessWhenPredictionNodeEvent()
...
+    @Test
+    void shouldProcessWhenPredictionIsConfirmed()
+    {
+        String parentId = "parent_id";
+        String descriptionProperty = "cm:description";
+        String predictedDescription = "predicted description";
+
+        given(mockEvent.getData()).willReturn(mock());
+        given(mockEvent.getData().getResource()).willReturn(mock());
+        given(mockEvent.getData().getResource().getNodeType()).willReturn(PREDICTION_NODE_TYPE);
```
tpage-alfresco, line 314: This is a neat way to create nested mocks :+1:

Pawel-608: yep super smart🤯 I think Krystian or Maciek invented that

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/EventProcessor.java
Pawel-608: yep, you're right, I deleted this check

tpage-alfresco (approved):
LGTM - thanks Pawal!

PR #358, opened 2024-05-21
[ACS-7949] Add new endpoint to confirm/reject predictions for node properties

MichalKinas:
PR adding new endpoint to either confirm or reject prediction for node properties.

### DIFF
--- a/prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/exception/PredictionStateChangedException.java
+++ b/prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/exception/PredictionStateChangedException.java
@@ -25,7 +25,9 @@
  */
 package org.alfresco.hxi_connector.prediction_applier.rest.api.exception;
 
-public class PredictionStateChangedException extends RuntimeException
+import org.alfresco.error.AlfrescoRuntimeException;
+
+public class PredictionStateChangedException extends AlfrescoRuntimeException
 {
     public PredictionStateChangedException(String message)
     {
--- a/prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/exception/PredictionStateChangedException.java
+++ b/prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/exception/PredictionStateChangedException.java
@@ -1,0 +1,34 @@
+/*-
+ * #%L
+ * Alfresco HX Insight Connector
+ * %%
+ * Copyright (C) 2023 - 2024 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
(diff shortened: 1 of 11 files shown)


prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/service/PredictionServiceImpl.java
```diff
@@ -160,4 +165,48 @@ public List<String> getPredictedProperties(NodeRef nodeRef)
...
+        Prediction prediction = childAssocToPrediction(parentAssocRef);
+        QName propertyQName = QName.createQName(prediction.getProperty(), namespaceService);
+        if (existingProperties.get(propertyQName).equals(prediction.getPredictionValue()))
+        {
+            updatePredictionState(predictionNodeRef, parentNode, existingProperties, prediction, propertyQName, reviewStatus);
+        }
+        else
+        {
+            throw new PredictionStateChangedException(prediction.getProperty() + " property value has changed, prediction is no longer valid!");
+        }
```
Pawel-608, line 184: idk if everyone will agree with me, but I prefer not to have nested if-else structures, so I would suggest doing something like this:
```
if (!existingProperties.get(propertyQName).equals(prediction.getPredictionValue()))
{
throw new PredictionStateChangedException(prediction.getProperty() + " property value has changed, prediction is no longer valid!");
}

updatePredictionState(predictionNodeRef, parentNode, existingProperties, prediction, propertyQName, reviewStatus);
```

If you do this, you can also get rid of the `updatePredictionState` method and try to simplify it in a similar way

mpichura: I agree with Pawel - you should try to get rid of those nested conditionals.

MichalKinas: Good idea, refactored

Pawel-608 (approved):
LGTM

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/exception/PredictionStateChangedException.java
```diff
@@ -0,0 +1,34 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.hxi_connector.prediction_applier.rest.api.exception;
+
+public class PredictionStateChangedException extends RuntimeException
```
tpage-alfresco, line 28: Can we extend `AlfrescoRuntimeException` instead? It has some capabilities for localisation.

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/service/PredictionServiceImpl.java
```diff
@@ -160,4 +165,34 @@ public List<String> getPredictedProperties(NodeRef nodeRef)
                 .map(qName -> qName.toPrefixString(namespaceService))
                 .collect(toList());
     }
+
+    @Override
+    public void reviewPrediction(NodeRef predictionNodeRef, ReviewStatus reviewStatus) throws PredictionStateChangedException
+    {
+        ChildAssociationRef parentAssocRef = nodeService.getPrimaryParent(predictionNodeRef);
+        NodeRef parentNode = parentAssocRef.getParentRef();
+        Map<QName, Serializable> existingProperties = new HashMap<>(nodeService.getProperties(parentNode));
```
tpage-alfresco, line 174: I think it would be more efficient to work with a single property here (assuming the API only allows confirming/rejecting a single property).

tpage-alfresco (approved):
On the whole looks really good @MichalKinas

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/exception/PredictionStateChangedException.java
MichalKinas: Sure, changed

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/service/PredictionServiceImpl.java
MichalKinas: Good point, changed

PR #367, opened 2024-05-28
ACS-7586: mark processed batches as completed

Pawel-608:


prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/hx_insight/PredictionCollector.java
```diff
@@ -97,64 +86,65 @@ public void configure()
...
+            .loopDoWhile(statusCodeNot204())
+                .toD(predictionsUrl)
+                .choice().when(statusCodeNot204())
+                    .log(TRACE, log, "Processing page ${headers.%s} of predictions in batch ${headers.%s}, ${body}, ${header.CamelHttpResponseCode}".formatted(PREDICTIONS_PAGE_NO_HEADER, BATCH_ID_HEADER))
+                    .unmarshal(predictionsDataFormat)
+                    .split(body())
+                        .marshal(predictionDataFormat)
+                        .to(insightPredictionsProperties.bufferEndpoint())
                     .end()
+                    .setBody(simple("{\"status\": \"COMPLETE\", \"currentPage\": ${headers.%s}}".formatted(PREDICTIONS_PAGE_NO_HEADER)))
```
tpage-alfresco, line 135: I think (but am not sure) that the status shouldn't be `COMPLETE` until the last page has been processed. Let's check on the call in 15 mins.

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/hx_insight/PredictionCollector.java
```diff
@@ -97,64 +86,65 @@ public void configure()
...
+            .setHeader(BATCH_ID_HEADER, simple("${body.id}"))
+            .setHeader(PREDICTIONS_PAGE_NO_HEADER, constant(1))
+            .loopDoWhile(statusCodeNot204())
+                .toD(predictionsUrl)
+                .choice().when(statusCodeNot204())
+                    .log(TRACE, log, "Processing page ${headers.%s} of predictions in batch ${headers.%s}, ${body}, ${header.CamelHttpResponseCode}".formatted(PREDICTIONS_PAGE_NO_HEADER, BATCH_ID_HEADER))
+                    .unmarshal(predictionsDataFormat)
+                    .split(body())
+                        .marshal(predictionDataFormat)
+                        .to(insightPredictionsProperties.bufferEndpoint())
```
tpage-alfresco, line 133: Not a change for this PR, but I don't think the queue name should be configurable. We should review all the properties and see what makes sense to ask the user to set.

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/model/prediction/PredictionBatch.java
```diff
@@ -25,20 +25,10 @@
...
 public record PredictionBatch(
-        @JsonProperty("_id") String id,
-        String modelId,
-        String fieldConfigurationId,
-        String field,
-        String enrichmentType,
-        Double threshold,
-        @JsonProperty("dateCreated") String creationDate,
-        String status,
-        Integer currentPage,
```
tpage-alfresco, line 41: I would have expected us to need the current page (e.g. if restarting the application). Does it work to just ask HxI for the latest page?
Don't we need the other properties? (e.g. `field` seems essential and we store `enrichmentType` in the ACS content model).

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/hx_insight/PredictionCollector.java
tpage-alfresco: The HxI team are going to get back to us about the set of valid values for this field. We're expecting it to do something like:
```
APPROVED -> IN PROGRESS -> COMPLETE
APPROVED: Ready to start
IN PROGRESS: Not all pages in AMQ
COMPLETE: All pages in AMQ
```

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/model/prediction/PredictionBatch.java
Pawel-608: I think all data we need is also stored (field, enrichmentType) in individual prediction entry, so we don't use these fields. Also "currentPage" property doesn't tell us anything as we don't have number of all pages🤦‍♂️ - so we cannot resume processing if app will fail because the only info we have about every batch is: status "COMPLETE", page: 5, but we don't know if there is just 5 pages or there's more of them

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/hx_insight/PredictionCollector.java
tpage-alfresco: Here's the message from Ryan: https://alfresco.slack.com/archives/C057YBK8H8D/p1716904597681529

Pawel-608: yep, you're right, I also confirmed with Ryan, fixed

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/model/prediction/PredictionBatch.java
tpage-alfresco: I thought that we would be able to see:
* IN_PROGRESS, page 5: We need to process page 5
* COMPLETE, page 5: We have processed all pages (and the last page was page 5, or maybe 4)

If the application goes down part way through and we see IN_PROGRESS, page 5 then we may accidentally process the same page twice, but at least we shouldn't lose data like this.

tpage-alfresco:
We should have a test for the IN_PROGRESS -> COMPLETE transistion.

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/model/prediction/PredictionBatch.java
Pawel-608: yep, you're right, I restored currentPage property, I think status is not needed as we always want to query for batches in some status so we already know it

tpage-alfresco (approved):
Test is coming in separate PR: https://github.com/Alfresco/hxinsight-connector/pull/382

PR #382, opened 2024-06-02
ACS-7585: test if batch status is updated

Pawel-608:


prediction-applier/src/integration-test/java/org/alfresco/hxi_connector/prediction_applier/domain/usecase/e2e/PredictionApplicationIntegrationTest.java
```diff
@@ -33,18 +33,22 @@
...
         // given
-        containerSupport.prepareHxInsightToReturnPredictionBatch(NODE_ID, PREDICTED_VALUE);
+        containerSupport.prepareHxInsightToReturnPredictionBatch(BATCH_ID, NODE_ID, PREDICTED_VALUE);
 
         // when
         triggerPredictionsCollection();
 
         // then
+        containerSupport.expectBatchStatusWasUpdated(BATCH_ID, "IN_PROGRESS", 1);
+        containerSupport.expectBatchStatusWasUpdated(BATCH_ID, "COMPLETE", 2);
```
tpage-alfresco, line 50: Is it possible that the application gets to COMPLETE before noticing that it was IN_PROGRESS?

Pawel-608: no, it always goes like:
- get page
- mark batch as IN_PROGRESS + set currentPage to 1
- get next page
- mark batch as IN_PROGRESS + set currentPage to 2
- after processing all pages mark batch as COMPLETE

tpage-alfresco (approved):

PR #385, opened 2024-06-04
ACS-8067: fix compilation issue with ACS 7.3.x

Pawel-608:


tpage-alfresco (approved):
Nice find and fix :+1:

PR #386, opened 2024-06-04
ACS-8067: test if app is compatible with 7.3.x and 7.4.x ACS versions

Pawel-608:


tpage-alfresco (approved):
Well done for making it fail with 7.3 :rocket: 

Assuming this passes with your fix PR then this is good to merge.

PR #399, opened 2024-06-11
ACS-8115: add questions endpoint

Pawel-608:


prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/QuestionApi.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.alfresco.hxi_connector.prediction_applier.rest.api.model.Question;
+import org.alfresco.rest.framework.WebApiDescription;
+import org.alfresco.rest.framework.resource.EntityResource;
+import org.alfresco.rest.framework.resource.actions.interfaces.EntityResourceAction;
+import org.alfresco.rest.framework.resource.parameters.Parameters;
+
+@Slf4j
+@EntityResource(name = "questions", title = "Questions about documents")
+public class QuestionApi implements EntityResourceAction.Create<Question>
```
Pawel-608, line 41: Should I follow convention and call this class “QuestionEntityResource”? To me, these types of names are completely meaningless, so I would suggest renaming the other “controllers” in a similar way, so that their names actually mean something

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/QuestionApi.java
```diff
@@ -0,0 +1,52 @@
...
+{
+
+    @Override
+    @WebApiDescription(title = "Ask question", successStatus = Status.STATUS_OK)
+    public List<Question> create(List<Question> questions, Parameters parameters)
+    {
+        log.info("Received questions: {}", questions);
+
+        return List.of(Question.createResponse("questionId"));
```
Pawel-608, line 50: maybe someone knows if it is possible to have different request and response types?

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/QuestionApi.java
```diff
@@ -0,0 +1,52 @@
...
+import org.alfresco.rest.framework.resource.parameters.Parameters;
+
+@Slf4j
+@EntityResource(name = "questions", title = "Questions about documents")
+public class QuestionApi implements EntityResourceAction.Create<Question>
+{
+
+    @Override
+    @WebApiDescription(title = "Ask question", successStatus = Status.STATUS_OK)
+    public List<Question> create(List<Question> questions, Parameters parameters)
```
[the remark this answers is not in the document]
tpage-alfresco: I can't think of a use case for accepting more than one question at a time, so I suggest we set this as a limit.

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/QuestionApi.java
tpage-alfresco: For this API I think it's fine to just have the same type in the request and response. We just need to set the id in the response (as you've done).

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/model/Question.java
```diff
@@ -0,0 +1,49 @@
...
+import lombok.Setter;
+import lombok.ToString;
+
+@ToString
+@Setter(AccessLevel.PRIVATE)
+public class Question
+{
+    private String questionId;
+    @Getter
+    private String question;
```
tpage-alfresco, line 40: Can this be set?

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/model/Question.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public static Question createResponse(String questionId)
+    {
+        Question question = new Question();
+        question.setQuestionId(questionId);
```
tpage-alfresco, line 45: The standard format is to include all the information from the request, and then add some generated data.

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/QuestionApi.java
Pawel-608: okay, changed!

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/QuestionApi.java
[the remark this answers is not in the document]
Pawel-608: done - 400 is returned in case more than one question was asked

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/model/Question.java
Pawel-608: We do not need to set it as we won't construct this object manually

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/model/Question.java
Pawel-608: fixed!

krdabrowski (approved):
LGTM

PR #400, opened 2024-06-12
ACS-8119: add restriction query to Question

Pawel-608:


krdabrowski (approved):

tpage-alfresco (approved):
Ok - maybe that ticket was too small!

PR #405, opened 2024-06-13
Fix Veracode SCA scan

dsibilio:
Fixing the Veracode SCA scan once again! 🤞

**Context**: the extra single quotes in [this commit](https://github.com/Alfresco/hxinsight-connector/commit/37688eb127796c432abda35fb48cd14a4a7623ac) broke it, but I guess we didn't notice as it fails silently due to the `continue-on-error` option being `true`.

**Successful run after the change**: https://github.com/Alfresco/hxinsight-connector/actions/runs/9501397929/job/26186873098?pr=405 ✅


tpage-alfresco (approved):
Thanks!

PR #406, opened 2024-06-13
Improve local development instructions

dsibilio:
Also added the `distribution/src/main/resources/docker-compose/volume` to the `.gitignore` file as I don't think this is meant to be committed, yet when running the environment locally this seemed to be autogenerated and I guess we don't want people pushing their volumes by mistake.


tpage-alfresco (approved):
I have added the `volume` directory to my local `info/exclude` file. Not sure why I didn't just do this instead.

PR #412, opened 2024-06-18
ACS-8120: get rid of camel in common-authentication module

Pawel-608:


tpage-alfresco (approved):
The change looks a really nice simplification. If you can also get the CI to be happy with this then we should definitely go with this approach.

krdabrowski (approved):
LGTM

PR #415, opened 2024-06-19
Add deserialisation tests for PredictionModel.

tpage-alfresco:
Using the test structure introduced for the Question model object.


Pawel-608 (approved):
LGTM! one small remark - maybe it is worth checking if whole object is deserialized correctly, not just one field

krdabrowski (approved):

PR #424, opened 2024-06-24
ACS-6826 test to apply prediction to updated node

azakrzewski-hy:
[ACS-6826](https://hyland.atlassian.net/browse/ACS-6826)
Upload a new version of a document and check that HxI receives the metadata update then ACS receives prediction for the node.
Added test covers scenario:
1. node is created
2. prediction is generated for the node from step 1.
3. node from step 1. is updated together with field that has prediction from step 2.
4. prediction for updated node (in step 3.) is generated but updated node should not apply this prediction and keep updated user value

[ACS-6826]: https://hyland.atlassian.net/browse/ACS-6826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java
+++ b/e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java
@@ -159,21 +159,21 @@ public class UpdateNodeE2eTest
         @Cleanup
         InputStream fileContent = new ByteArrayInputStream(DUMMY_CONTENT.getBytes());
         Node createdNode = repositoryNodesClient.createNodeWithContent(PARENT_ID, "dummy2.txt", fileContent, "text/plain");
-        Node updateNode = repositoryNodesClient.updateNodeWithContent(createdNode.id(), UPDATE_NODE_PROPERTIES);
+        Node updatedNode = repositoryNodesClient.updateNodeWithContent(createdNode.id(), UPDATE_NODE_PROPERTIES);
         RetryUtils.retryWithBackoff(() -> verify(moreThanOrExactly(1), postRequestedFor(urlEqualTo("/ingestion-events"))
-                .withRequestBody(containing(updateNode.id()))));
+                .withRequestBody(containing(updatedNode.id()))));
         WireMock.reset();
-        prepareHxInsightMockToReturnPredictionFor(updateNode.id());
+        prepareHxInsightMockToReturnPredictionFor(updatedNode.id());
 
         // when
         WireMock.setScenarioState(LIST_PREDICTIONS_SCENARIO, PREDICTIONS_AVAILABLE_STATE);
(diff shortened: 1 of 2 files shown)


e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java
```diff
@@ -89,6 +89,14 @@ public class UpdateNodeE2eTest
               }
             ]
             """;
+    private static final String UPDATE_NODE_PROPERTIES = """
+            {
+                "properties": {
+                    "cm:versionLabel": "1.1",
+                    "cm:title": "User title"
+                    }
```
tpage-alfresco, line 97:
```suggestion
                }
```

azakrzewski-hy:
> The test looks fine. It looks like a very similar scenario to the existing test because the ACS update doesn't impact the prediction.

It is very similar to the existing test. Idea was to cover scenario for:
"Upload a new version of a document and check that HxI receives the metadata update"
If you say that there is no much value in adding this test we can drop it

krdabrowski:
> "Upload a new version of a document and check that HxI receives the metadata update"
Adam if you would like to verify only this part, then maybe it's worth considering removing the redundant part - prediction mock publication and test only what you meant?

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java
```diff
@@ -25,17 +25,7 @@
...
-import static com.github.tomakehurst.wiremock.client.WireMock.containing;
-import static com.github.tomakehurst.wiremock.client.WireMock.exactly;
-import static com.github.tomakehurst.wiremock.client.WireMock.get;
-import static com.github.tomakehurst.wiremock.client.WireMock.givenThat;
-import static com.github.tomakehurst.wiremock.client.WireMock.moreThanOrExactly;
-import static com.github.tomakehurst.wiremock.client.WireMock.postRequestedFor;
-import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
-import static com.github.tomakehurst.wiremock.client.WireMock.urlPathTemplate;
-import static com.github.tomakehurst.wiremock.client.WireMock.verify;
+import static com.github.tomakehurst.wiremock.client.WireMock.*;
```
krdabrowski, line 28: Avoid wildcard imports

krdabrowski (approved):
LGTM

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java
azakrzewski-hy: Changed wildcard import for separate method imports

PR #431, opened 2024-06-26
ACS-8278: Add retry to authentication in JAMP

krdabrowski:
https://hyland.atlassian.net/browse/ACS-8278


prediction-applier-extension/src/test/java/org/alfresco/hxi_connector/prediction_applier/client/HxInsightAuthClientTest.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @AfterAll
+    void afterAll()
+    {
+        rootLoggerConfig.removeAppender(testAppender.getName());
+    }
+
+    @Test
+    void authenticate()
```
mpichura, line 87:
```suggestion
    void givenAuthException_whenAuthenticate_thenFailAfterRetry()
```

mpichura (approved):
LGTM, I had one suggestion for test method naming.

common/src/main/java/org/alfresco/hxi_connector/common/config/properties/Retry.java
```diff
@@ -36,41 +38,56 @@
...
     private int attempts = RETRY_ATTEMPTS_DEFAULT;
     @PositiveOrZero
     private int initialDelay = RETRY_INITIAL_DELAY_DEFAULT;
     @Positive private double delayMultiplier = RETRY_DELAY_MULTIPLIER_DEFAULT;
-    @NotNull private Set<Class<? extends Throwable>> reasons = RETRY_REASONS;
+    @NotNull private Set<Class<? extends Throwable>> reasons;
+
+    public Retry()
+    {
+        this.reasons = Stream.concat(RETRY_REASONS_BASIC.stream(), Stream.of(HttpHostConnectException.class, NoHttpResponseException.class, MalformedChunkCodingException.class))
```
tpage-alfresco, line 72: I don't understand why you moved these out of `RETRY_REASONS_BASIC`? Is this in preparation for another PR and you will make `RETRY_REASONS_BASIC` public in that?

common/src/main/java/org/alfresco/hxi_connector/common/config/properties/Retry.java
```diff
@@ -36,41 +38,56 @@
...
+                .collect(Collectors.toSet());
+    }
+
+    public Retry(int attempts, int initialDelay, double delayMultiplier)
+    {
+        this();
```
tpage-alfresco, line 78: You could change the order of the constructors so that the default constructor calls the 3-arg and the 3-arg calls the 4-arg.

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/client/HxInsightAuthClient.java
```diff
@@ -0,0 +1,93 @@
...
+        long delay = retryProperties.initialDelay();
+        while (true)
+        {
+            try
+            {
+                return supplier.get();
+            }
+            catch (Exception e)
+            {
+                if (CollectionUtils.isNotEmpty(retryProperties.reasons()) && retryProperties.reasons().contains(e.getClass()))
```
tpage-alfresco, line 67:
```suggestion
                if (retryProperties.reasons() != null && retryProperties.reasons().contains(e.getClass()))
```
or you can do this with `Optional.ofNullable`.

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/client/HxInsightAuthClient.java
```diff
@@ -0,0 +1,93 @@
...
+                        throw e;
+                    }
+                    log.info("Attempt {} of {} failed, retrying after {}ms", attempt, maxAttempts, delay);
+                    try
+                    {
+                        TimeUnit.MILLISECONDS.sleep(delay);
+                    }
+                    catch (InterruptedException ex)
+                    {
+                        log.warn("Cannot pause retryable operation due to InterruptedException: %s".formatted(ex.getMessage()), e);
```
tpage-alfresco, line 82:
```suggestion
                        log.atWarn().log("Cannot pause retryable operation due to InterruptedException: {}", ex.getMessage(), e);
```

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/client/HxInsightAuthClient.java
```diff
@@ -0,0 +1,93 @@
...
+                    }
+                    delay = Math.round(delay * retryProperties.delayMultiplier());
```
tpage-alfresco, line 84: How about making `delay` a double and moving the `Math.round` to the line that calls `sleep`?

tpage-alfresco (approved):
Code looks fine. I added a few comments/questions.

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/client/HxInsightAuthClient.java
krdabrowski: Both solutions are OK IMO - as you mentioned rounding error is invisible in this row of numbers. In the other hand I like your hint so I will fix it 😅

common/src/main/java/org/alfresco/hxi_connector/common/config/properties/Retry.java
krdabrowski: I've excluded those exceptions from `RETRY_REASONS_BASIC` to be able to use this class in JAMP and avoid including `httpcomponents` dependency into jar

common/src/main/java/org/alfresco/hxi_connector/common/config/properties/Retry.java
krdabrowski: Good hint! I will fix it

PR #432, opened 2024-06-26
Remove sping-boot-starter-test from parent pom

Pawel-608:


common-test/pom.xml
```diff
@@ -18,6 +18,16 @@
             <artifactId>logback-classic</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <exclusions>
```
tpage-alfresco, line 24: We could add this exclusion via `dependencyManagement` in the root pom.

tpage-alfresco (approved):
LGTM :+1:

krdabrowski (approved):

PR #433, opened 2024-06-26
ACS-8138 Wire up agents endpoint to Hx Insight API.

tpage-alfresco:


tpage-alfresco:
This needs [feature/ACS-8121-call-hxi-api](https://github.com/Alfresco/hxinsight-connector/tree/feature/ACS-8121-call-hxi-api) to be complete. Currently testing with docker-compose is failing with:
```
Cannot invoke "org.alfresco.hxi_connector.common.config.properties.Retry.reasons()" because the return value of "org.alfresco.hxi_connector.common.adapters.auth.config.properties.AuthProperties.getRetry()" is null
```
which is the same as for question submission.

krdabrowski (approved):
LGTM

PR #434, opened 2024-06-26
Bye bye Spring XML-config

krdabrowski:


prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/rest/api/NodeEntityResource.java
```diff
@@ -43,13 +44,21 @@
...
 public class NodeEntityResource implements EntityResourceAction.ReadById<NodeWithPrediction>
 {
-    private NodeService nodeService;
-    private NodesImpl nodes;
-    private PredictionService predictionService;
+    private final NodeService nodeService;
+    private final NodesImpl nodes;
+    private final PredictionService predictionService;
+
+    public NodeEntityResource(@Qualifier("NodeService") NodeService nodeService, NodesImpl nodes,
```
Pawel-608, line 55: do you have to add qualifier?

Pawel-608 (commented):
looks great!

prediction-applier-extension/src/main/java/org/alfresco/hxi_connector/prediction_applier/config/ExtensionConfiguration.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Bean
+    public AlwaysProceedMethodInterceptor predictionServiceSecurity()
+    {
+        return new AlwaysProceedMethodInterceptor();
+    }
+
+    @Bean
+    @Primary
+    public PredictionService predictionService(BeanFactory beanFactory, @Qualifier("predictionServiceImpl") PredictionService predictionService)
```
tpage-alfresco, line 72: Assuming I read this right then I think the qualifier should be `"predictionService"`

krdabrowski: Nope, because this service is declared as:
```
@Service
public class PredictionServiceImpl implements PredictionService
```
so the default id for this bean will be: `predictionServiceImpl`

tpage-alfresco (changes requested):
After discussion then we concluded that although this is really clever the world is not ready for Spring annotations yet :face_exhaling:

krdabrowski:
As @tpage-alfresco mentioned within team we've decided to not merge this PR and stay with XML-config 😮‍💨  Closing the PR.

PR #447, opened 2024-07-01
ACS-8300: add hxi client tests

Pawel-608:


hxinsight-extension/src/integration-test/java/org/alfresco/hxi_connector/hxi_extension/rest/api/QuestionsEntityResourceIntegrationTest.java
```diff
@@ -0,0 +1,65 @@
...
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(locations = {"classpath:rest/api/question-entity-resource-test-context.xml"})
+public class QuestionsEntityResourceIntegrationTest
+{
+
+    @Autowired
+    private QuestionsEntityResource questionsEntityResource;
+
+    @Test
+    public void shouldFailIfAskedMultipleQuestions()
```
mpichura, line 54: I know this might be tricky in JAMP, but did you try to write a test for a positive path?

hxinsight-extension/src/integration-test/resources/rest/api/question-entity-resource-test-context.xml
```diff
@@ -0,0 +1,66 @@
...
+  ~ Alfresco is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public License
+  ~ along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+  ~ #L%
+  -->
+<beans xmlns="http://www.springframework.org/schema/beans"
```
mpichura, line 27: WDYT about having this test config as Java annotation based config?
I think it would increase the readability and this case (tests) could be considered as not breaking the common repo/AMPs convention.
I am okay with XML config - that is just an idea we could discuss.

mpichura (commented):
I think we should have the positive path tested as well.

hxinsight-extension/src/integration-test/java/org/alfresco/hxi_connector/hxi_extension/rest/api/QuestionsEntityResourceIntegrationTest.java
Pawel-608: I removed Integration test from this pr - I'll add it later

hxinsight-extension/src/integration-test/resources/rest/api/question-entity-resource-test-context.xml
Pawel-608: I removed Integration test from this pr - I'll add it later

PR #451, opened 2024-07-01
ACS-8262 Fix E2E test failures

mpichura:
This PR splits E2E test class `CreateNodeE2eTest` into a parent abstract class `CreateNodeE2eTestBase` and two child classes: `CreateNodeE2eTest`, `CreateNodeDCE2eTest`.
GHA pipeline will only run `CreateNodeDCE2eTest` in a newly step: `Start docker-compose and run e2e docker-compose tests` which will also spin up a minimal docker-compose environment from command line.
On our local machines, we'll still run `CreateNodeE2eTest` based on `testcontainers`. 
Above test classes are executed conditionally basing on environment variable named `GHA_RUN_DC` being set in the job definition.


Pawel-608 (approved):
lgtm

.github/workflows/ci.yml
```diff
@@ -148,7 +148,18 @@ jobs:
...
+          GHA_RUN_DC: true
+        run: |
+          ./scripts/ci/start-compose.sh distribution/src/main/resources/docker-compose/docker-compose-minimal.yml
+          mvn ${{ env.MAVEN_CLI_OPTS }} test -pl 'prediction-applier-extension,e2e-test' -am -Dtest=CreateNodeDCE2eTest -DfailIfNoTests=false -Dalfresco-platform.version=${{ matrix.repoVersion }}
+      - name: "Stop docker-compose"
+        if: always()
+        run: ./scripts/ci/stop-compose.sh distribution/src/main/resources/docker-compose/docker-compose-minimal.yml
       - name: "Run e2e tests"
+        env:
+          GHA_RUN_DC: false
```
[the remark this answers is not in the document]
mpichura: I thought that when I set it in a previous step its scope will be limited to that particular step but somehow it preserved and made this step run the docker-compose tests which run when this variable is set to `true`. Thus, I am setting that to `false` although I would like to have this variable unset.

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/CreateNodeDCE2eTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @AfterEach
+    void tearDown()
+    {
+        WireMock.reset();
+    }
+
+    @Test
+    @SuppressWarnings({"PMD.JUnitTestsShouldIncludeAssert"})
+    void testCreateNodeContainingImageFile() throws IOException
```
krdabrowski, line 65: I think both test methods entirely could be moved to parent class 🤔

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/CreateNodeE2eTest.java
```diff
@@ -120,61 +103,14 @@ void tearDown()
...
-        // then
-        RetryUtils.retryWithBackoff(() -> {
-            List<S3Object> actualBucketContent = awsS3Client.listS3Content();
-            assertThat(actualBucketContent.size()).isEqualTo(initialBucketContent.size() + 1);
-
-            WireMock.verify(exactly(1), postRequestedFor(urlEqualTo("/presigned-urls")));
-            WireMock.verify(moreThanOrExactly(2), postRequestedFor(urlEqualTo("/ingestion-events"))
-                    .withRequestBody(containing(createdNode.id())));
-        }, INITIAL_DELAY_MS);
+        testCreateNodeContainingImageFileBase();
```
krdabrowski, line 106: Here the same - both test methods could be entirely placed in parent class IMO

krdabrowski (approved):
Overall LGTM, I have only few remarks

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/CreateNodeE2eTest.java
mpichura: Yep - thanks for catching that - I totally had that in mind when annotated the classes with `@DisabledIfEnvironmentVariable`/`@EnabledIfEnvironmentVariable`. 
At first try methods were annotated as I thought it wouldn't work on the whole class - now it's definitely worth moving to the parent.

.github/workflows/ci.yml
```diff
@@ -148,7 +148,18 @@ jobs:
...
+      - name: "Start docker-compose and run e2e docker-compose tests"
+        env:
+          GHA_RUN_DC: true
+        run: |
+          ./scripts/ci/start-compose.sh distribution/src/main/resources/docker-compose/docker-compose-minimal.yml
+          mvn ${{ env.MAVEN_CLI_OPTS }} test -pl 'prediction-applier-extension,e2e-test' -am -Dtest=CreateNodeDCE2eTest -DfailIfNoTests=false -Dalfresco-platform.version=${{ matrix.repoVersion }}
+      - name: "Stop docker-compose"
+        if: always()
+        run: ./scripts/ci/stop-compose.sh distribution/src/main/resources/docker-compose/docker-compose-minimal.yml
       - name: "Run e2e tests"
```
tpage-alfresco, line 160: We might want to split out `run_e2e_test` into two jobs `run_docker_e2e_test` and `run_tc_e2e_test`. No need for this to be part of this PR, but I wanted to mention it while I thought of it.

.github/workflows/ci.yml
```diff
@@ -148,7 +148,18 @@ jobs:
         run: mvn ${{ env.MAVEN_CLI_OPTS }} clean install -DskipTests -Dalfresco-platform.version=${{ matrix.repoVersion }}
       - name: "Build docker images"
         run: bash ./scripts/ci/buildDockerImages.sh
+      - name: "Start docker-compose and run e2e docker-compose tests"
+        env:
+          GHA_RUN_DC: true
+        run: |
+          ./scripts/ci/start-compose.sh distribution/src/main/resources/docker-compose/docker-compose-minimal.yml
+          mvn ${{ env.MAVEN_CLI_OPTS }} test -pl 'prediction-applier-extension,e2e-test' -am -Dtest=CreateNodeDCE2eTest -DfailIfNoTests=false -Dalfresco-platform.version=${{ matrix.repoVersion }}
```
tpage-alfresco, line 156: Can we make this more general with a wildcard? `*DCE2eTest` or something like that?

.github/workflows/ci.yml
mpichura: We could do that but in both jobs we'd need to fully build the application and docker images which is the longest part of them (I know it would be parallelized but it still is some duplication that could be avoided).
Apart from that, I like the idea as it seems to be a bit clearer.
Maybe we could discuss that team-wise and implement in a separate PR if needed?

.github/workflows/ci.yml
mpichura: We can, but will we need it to be more general?

.github/workflows/ci.yml
tpage-alfresco: Discussed and this is not desired since the tests take a lot less time than building the images. We might do this if necessary in the future.

.github/workflows/ci.yml
tpage-alfresco: Currently only this test is suffering from this issue, so we don't need to generalise. When we add more d-c e2e tests then we can revisit this.

PR #456, opened 2024-07-03
ACS-8122: add error handling to hxi api call

Pawel-608:


hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
```diff
@@ -71,14 +75,15 @@ public String askQuestion(Question question)
...
+            ensureThat(httpResponse.statusCode() == EXPECTED_STATUS_CODE,
+                    () -> new WebScriptException(httpResponse.statusCode(), "Request to hxi failed"));
 
             return objectMapper.readValue(httpResponse.body(), QuestionResponse.class)
                     .questionId();
         }
         catch (IOException | InterruptedException e)
         {
-            throw new HxInsightConnectorRuntimeException("Failed to ask question", e);
+            throw new WebScriptException(INTERNAL_SERVER_ERROR, "Failed to ask question", e);
```
tpage-alfresco, line 86: I think we want something that's more obviously retriable - like 502 Service Unavailable or 503 Bad Gateway.

tpage-alfresco: Since we only have one client for this private API then maybe we should just ask @MichalKinas what status codes they retry on.

hxinsight-extension/src/main/resources/alfresco/module/alfresco-hxinsight-connector-hxinsight-extension/context/service-context.xml
```diff
@@ -80,6 +80,7 @@
         <property name="exceptionMappings">
             <map merge="true">
                 <entry key="org.alfresco.hxi_connector.hxi_extension.rest.api.exception.PredictionStateChangedException" value="#{T(org.springframework.extensions.webscripts.Status).STATUS_CONFLICT}" />
+                <entry key="org.alfresco.hxi_connector.common.exception.HxInsightConnectorRuntimeException" value="#{T(org.springframework.extensions.webscripts.Status).STATUS_INTERNAL_SERVER_ERROR}" />
```
krdabrowski, line 83: In the client you are throwing `WebScriptException` and here you have an declaration for `HxInsightConnectorRuntimeException`... this is correct? 🤔

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
mpichura: 503 Service Unavailable would fit here best IMO.

Pawel-608: I asked @AleksanderSklorz and he told me that we should just pass status code to them, so I'll leave it as it is

AleksanderSklorz: I have one question for clarification - which status code will be used to let us know we should still retry call and which that we should stop retrying because there is something wrong on backend? Just to avoid situation that we will be doing infinite number of calls

tpage-alfresco: I think you should change this one because it's returning a 500 for a network issue. I don't think ADW will retry on a 500, but hopefully they do with a 503.

mpichura: We may also want to think about a generalized error message so that it can be i18-ed.

mpichura (approved):
LGTM

PR #461, opened 2024-07-04
Add all jars and license files to distribution zip

dsibilio:
While working on implementing the SAST scan on PR, I noticed that the distribution zip did not contain all that I was expecting. In particular, it only contained the jar and license files for the `live-ingester` module.

I talked with @tpage-alfresco and he confirmed that this was not the expected result, so I went ahead and added all other jars and license files to the distribution zip. This will also be useful for the SAST scan on PR feature which I'll continue working on a separate branch.

### Before

image

### After

image


tpage-alfresco (approved):
Thanks for the fix!

mpichura (approved):
Well done and thank you 👍

damianujma (approved):

PR #464, opened 2024-07-04
Unify error handling in HxInsightClient

Pawel-608:


tpage-alfresco (commented):
I'm not completely sure why you're proposing this change. I guess it's because `throwExceptionOnUnexpectedStatusCode` doesn't preserve the original status code? It might be neater to add an extra helper method to `ErrorUtils` (or somewhere else) that does the assert/try/catch logic.

tpage-alfresco (approved):
Great :+1:

PR #465, opened 2024-07-04
Remove unused Agent class

Pawel-608:


mpichura (approved):
Nice catch.

tpage-alfresco (approved):
Ah - well spotted that the class is not used! There is also the `Agent` class for communication with HxI, and I guess I got confused between these.

PR #466, opened 2024-07-05
ACS-8139 Expect agent id from ADW and pass it to HxI.

tpage-alfresco:


hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/model/QuestionModel.java
```diff
@@ -52,6 +52,8 @@ public class QuestionModel
     @NotBlank
     private String question;
     @NotBlank
+    private String agentId;
+    @NotBlank
```
mpichura, line 56: Will those validation annotations work in JAMP?

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/model/Question.java
```diff
@@ -38,5 +38,7 @@ public class Question
     @NotBlank
     private final String question;
     @NotBlank
+    private final String agentId;
+    @NotBlank
```
mpichura, line 42: Will those validation annotations work in JAMP?

Pawel-608 (approved):
LGTM

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/model/QuestionModel.java
tpage-alfresco: No - I raised a second PR immediately afterwards to take them out again!

PR #467, opened 2024-07-05
ACS-8139 Remove validation from JAMP objects.

tpage-alfresco:
These aren't currently doing anything and the validation will be performed by HxI anyway.


mpichura (approved):
Good catch. Are we planning to replace that with something we use in the repo?

tpage-alfresco:
I think most validation in the Repo is explicit - either in constructors or the service layer. This API is a bit different in that there isn't a service layer (it bypasses the DB completely) and validation is done by HxI.

If someone wants to investigate a neat way to add validation then great, otherwise I think it's fine to skip it here.

Pawel-608 (commented):
I'll take a look at that as maybe we can make validation in constructors, and it might be useful to have it in case we have to map something before passing to hxi

tpage-alfresco:
The only method to add validation in the constructors that I found was replacing all the Lombok constructors with explicit ones. The neatest solution I came up with was basically a set of statements like:
```
if (isBlank(question)) {
    throw new IllegalArgumentException(...);
}
```

PR #473, opened 2024-07-09
ACS-8152: Small E2E tests refactor - extracting common code.

mpichura:


common-test/src/main/java/org/alfresco/hxi_connector/common/test/docker/util/DockerContainers.java
```diff
@@ -107,27 +131,28 @@ public static String concatJavaOpts(String... opts)
...
-                -Xms1500m -Xmx1500m
-                """.formatted(
+        return MINIMAL_REPO_JAVA_OPTS.formatted(
                 postgresContainer.getUsername(),
                 postgresContainer.getPassword(),
                 postgresContainer.getNetworkAliases().stream().findFirst().get(),
                 postgresContainer.getDatabaseName(),
-                activemqContainer.getNetworkAliases().stream().findFirst().get())
+                activemqContainer.getNetworkAliases().stream().findFirst().get(),
+                "false")
```
Pawel-608, line 140: could you maybe create a helper method that will take all these arguments and describe them (it is hard to understand what this "true"/"false" is)

something like

getMinimalJavaOpts(postgresPass, postgresUname, postgres..., areTransformServicesEnabled)

Pawel-608: but maybe thats an overkill🤔 just throwing the idea

Pawel-608 (approved):
LGTM, just one suggestion

common-test/src/main/java/org/alfresco/hxi_connector/common/test/docker/util/DockerContainers.java
mpichura: I was thinking about that too but am afraid it may be an overkill. 
Method names are telling quite enough for me.

tpage-alfresco: Except for `-Xms1500m -Xmx1500m` this is just a map. Maybe we could reframe it that way to also address Pawel's suggestion?
I don't think I'm likely to ever look at this code again more than a handful of times, so maybe it is overkill though :smile:

tpage-alfresco (approved):

PR #485, opened 2024-07-15
Unify approach to base-url in live-ingester and prediction-applier (hardcode /v1 in both cases)

krdabrowski:


krdabrowski:
This is 1st option - the other one is to put `/v1` into env variable also in `prediction-applier` (unhardcode it) 🤔 Second option seems to be better

krdabrowski:
Closing in favor of: https://github.com/Alfresco/hxinsight-connector/pull/488

PR #488, opened 2024-07-16
Unify approach to base-url in live-ingester and prediction-applier (un-hardcode /v1 in both cases)

krdabrowski:


tpage-alfresco (approved):
LGTM

azakrzewski-hy (approved):

PR #489, opened 2024-07-16
ACS-8154: check users permissions e2e

Pawel-608:


e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/QuestionsPermissionsE2eTest.java
```diff
@@ -0,0 +1,178 @@
...
+                "test private file",
+                new ByteArrayInputStream("test file content".getBytes()),
+                "text/plain").id();
+    }
+
+    @Test
+    void adminShouldBeAbleToAskQuestionAboutPublicDocument()
+    {
+        // when
+        Response response = given().auth().preemptive().basic(ADMIN_USER.username(), ADMIN_USER.password())
```
krdabrowski, line 117: Maybe it's worth considering creating in `RepositoryClient` methods like `askQuestion(question)` and `askQuestionAs(question, user)` and move the code out from tests - WDYT?

Pawel-608: I don't like it tbh, I would prefer to stay as explicit as possible even if this means some repetitions

krdabrowski: I'm fine with that

PR #492, opened 2024-07-17
ACS-8351 Create agents avatar endpoint with mocked response.

aonych:
As per title.

### DIFF
--- a/hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
+++ b/hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
@@ -26,6 +26,7 @@
 package org.alfresco.hxi_connector.hxi_extension.rest.api;
 
 import java.io.File;
+import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.net.http.HttpClient;
@@ -58,6 +59,7 @@ public class AvatarRelation implements RelationshipResourceAction.ReadById<Binar
     {
         if (!avatarId.equals("-default-"))
         {
+            log.info("Avatar id is different than -default-");
             throw new NotFoundException(String.format("Avatar with id=%s not found", avatarId));
         }
 
@@ -77,9 +79,15 @@ public class AvatarRelation implements RelationshipResourceAction.ReadById<Binar
             File file = TempFileProvider.createTempFile(httpResponse.body(), "RenditionsApi-", ".png");
             return new FileBinaryResource(file, null);
         }
+        catch (IOException | InterruptedException e)
+        {
+            log.error("Error getting avatar image.", e);
+            throw new NotFoundException("Avatar image not found.");
+        }
         catch (Exception e)
         {
-            throw new RuntimeException(e);
+            log.error("Failed to create temp file.", e);
+            throw new RuntimeException("Failed to create avatar temp file.", e);
         }
     }
 
--- a/hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
+++ b/hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
@@ -36,6 +36,7 @@ import java.net.http.HttpResponse;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
+import org.alfresco.hxi_connector.hxi_extension.rest.api.exception.AgentAvatarException;
 import org.alfresco.rest.framework.WebApiDescription;
 import org.alfresco.rest.framework.core.exceptions.NotFoundException;
 import org.alfresco.rest.framework.core.exceptions.RelationshipResourceNotFoundException;
@@ -87,7 +88,7 @@ public class AvatarRelation implements RelationshipResourceAction.ReadById<Binar
         catch (Exception e)
         {
             log.error("Failed to create temp file.", e);
-            throw new RuntimeException("Failed to create avatar temp file.", e);
+            throw new AgentAvatarException("Failed to create avatar temp file.");
         }
     }
 
--- a/hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
+++ b/hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
@@ -1,0 +1,86 @@
+/*
+ * #%L
+ * Alfresco HX Insight Connector
+ * %%
+ * Copyright (C) 2023 - 2024 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.hxi_connector.hxi_extension.rest.api;
+
+import java.io.File;
+import java.io.InputStream;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+import org.alfresco.rest.framework.WebApiDescription;
+import org.alfresco.rest.framework.core.exceptions.NotFoundException;
+import org.alfresco.rest.framework.core.exceptions.RelationshipResourceNotFoundException;
+import org.alfresco.rest.framework.resource.RelationshipResource;
+import org.alfresco.rest.framework.resource.actions.interfaces.RelationshipResourceAction;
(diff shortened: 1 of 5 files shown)


hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/config/HxInsightClientConfig.java
```diff
@@ -42,5 +43,6 @@ public HxInsightClientConfig(@NotBlank String baseUrl)
         this.agentUrl = baseUrl + "/v1/agents";
         this.questionUrl = baseUrl + "/v1/questions";
         this.answerUrl = questionUrl + "/%s/answer";
+        this.avatarUrl = agentUrl + "/%s/avatars";
```
tpage-alfresco, line 46: Given this was previously set differently then is it actually being used? Can we just delete this entry?

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
```diff
@@ -0,0 +1,86 @@
...
+import org.alfresco.rest.framework.core.exceptions.NotFoundException;
+import org.alfresco.rest.framework.core.exceptions.RelationshipResourceNotFoundException;
+import org.alfresco.rest.framework.resource.RelationshipResource;
+import org.alfresco.rest.framework.resource.actions.interfaces.RelationshipResourceAction;
+import org.alfresco.rest.framework.resource.content.BinaryResource;
+import org.alfresco.rest.framework.resource.content.FileBinaryResource;
+import org.alfresco.rest.framework.resource.parameters.Parameters;
+import org.alfresco.util.TempFileProvider;
+
+@Slf4j
```
tpage-alfresco, line 48: There's currently no logging that I can see. I suggest adding something at trace or debug.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
```diff
@@ -0,0 +1,86 @@
...
+                .GET()
+                .build();
+
+        try
+        {
+            HttpResponse<InputStream> httpResponse = client.send(request, HttpResponse.BodyHandlers.ofInputStream());
+            File file = TempFileProvider.createTempFile(httpResponse.body(), "RenditionsApi-", ".png");
+            return new FileBinaryResource(file, null);
+        }
+        catch (Exception e)
```
tpage-alfresco, line 80: Please can we list the checked exceptions here. I assume they're all to do with IO? We can then convert the checked exceptions into the correct exceptions so that they return the right status code (e.g. in the future we will want to return a status code from Hx Insight).

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
```diff
@@ -0,0 +1,86 @@
...
+@RelationshipResource(name = "avatars", title = "Avatars of agents", entityResource = AgentsEntityResource.class)
+public class AvatarRelation implements RelationshipResourceAction.ReadById<BinaryResource>
+{
+    private HttpClient client;
+
+    @Override
+    @WebApiDescription(title = "Get Agent Avatar image")
+    public BinaryResource readById(String agentId, String avatarId, Parameters parameters) throws RelationshipResourceNotFoundException
+    {
+        if (!avatarId.equals("-default-"))
```
tpage-alfresco, line 59: There's not a huge amount to unit test in this PR, but I think this is something that we could add a test for.

tpage-alfresco (changes requested):
Great that you've been able to get up to speed with the project so quickly. A few changes to make and then I think we can move on to interacting with the Hx Insight API.

hxinsight-extension/src/test/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelationTest.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Test
+    public void shouldReturnAvatarForDefaultId()
+    {
+        // given
+        String avatarId = "-default-";
+
+        // when
+        BinaryResource binaryResource = avatarRelation.readById(AGENT_ID, avatarId, null);
```
tpage-alfresco, line 62: This unit test will fail if Wikipedia goes offline. Probably not very likely, but I don't like the test depending on an external resource like that. IMO it's fine to either delete the test for now, or extract the downloading code and use a mock in this test. I prefer the second option.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
```diff
@@ -0,0 +1,95 @@
...
+        try
+        {
+            HttpResponse<InputStream> httpResponse = client.send(request, HttpResponse.BodyHandlers.ofInputStream());
+            File file = TempFileProvider.createTempFile(httpResponse.body(), "RenditionsApi-", ".png");
+            return new FileBinaryResource(file, null);
+        }
+        catch (IOException | InterruptedException e)
+        {
+            log.error("Error getting avatar image.", e);
+            throw new AgentAvatarException("Avatar image cannot be received.", e);
```
tpage-alfresco, line 86: The last time we discussed this in the team we agreed upon 503 for this issue:
```
throw new WebScriptException(SC_SERVICE_UNAVAILABLE, "Failed to ask question", e);
```

I noticed that we have no consistency over it at the moment though, so I'll create a ticket to fix it in all endpoints.

tpage-alfresco: More generally it would be better to put this code in `HxInsightClient`. Have a look at what we're doing for the other endpoints.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
```diff
@@ -0,0 +1,95 @@
...
+        HttpRequest request = HttpRequest.newBuilder()
+                .uri(URI.create("https://www.wikipedia.org/portal/wikipedia.org/assets/img/Wikipedia-logo-v2@2x.png"))
+                .GET()
+                .build();
+
+        try
+        {
+            HttpResponse<InputStream> httpResponse = client.send(request, HttpResponse.BodyHandlers.ofInputStream());
+            File file = TempFileProvider.createTempFile(httpResponse.body(), "RenditionsApi-", ".png");
+            return new FileBinaryResource(file, null);
```
Pawel-608, line 81: is it necessary to first write response to file and then read it?

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
```diff
@@ -0,0 +1,95 @@
...
+    {
+        HttpRequest request = HttpRequest.newBuilder()
+                .uri(URI.create("https://www.wikipedia.org/portal/wikipedia.org/assets/img/Wikipedia-logo-v2@2x.png"))
+                .GET()
+                .build();
+
+        try
+        {
+            HttpResponse<InputStream> httpResponse = client.send(request, HttpResponse.BodyHandlers.ofInputStream());
+            File file = TempFileProvider.createTempFile(httpResponse.body(), "RenditionsApi-", ".png");
```
krdabrowski, line 80: Should we verify content type of received file, or we don't care and expect that HxI will always send a PNG? 🤔

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/config/HxInsightClientConfig.java
aonych: Deleted 👍

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
aonych: To be honest, this is just a sample method to test if the API will work with BinaryResource and if it will return an image instead of json. I want to get a green light from you first if the API looks good to start working on the logic part. 😃

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
aonych: Same as above 😄

krdabrowski: Ok, in this case LGTM

krdabrowski (approved):

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
Pawel-608: but we won't merge it like that🤞?

tpage-alfresco: I think this is fine to merge at the moment. It's going to need this code replacing and extracting to the `HxInsightClient` and we also need to add auth. Then we'll need an e2e test too.

tpage-alfresco (approved):
LGTM. @aonych Would you like me to create some subtasks for the ticket or are you happy you know what you need to do?

dsibilio (approved):

aonych:
> LGTM. @aonych Would you like me to create some subtasks for the ticket or are you happy you know what you need to do?

@tpage-alfresco It would be great!

hxinsight-extension/src/test/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelationTest.java
aonych: Updated.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/AvatarRelation.java
aonych: Updated

PR #493, opened 2024-07-17
Example of usage of MapStruct

krdabrowski:


tpage-alfresco (commented):
I'm ambivalent about this change. I like the removal of the mapping methods, but I'm concerned that we may occasionally want more complex mapping than simply moving data between differently named fields. (e.g. two fields being combined to populate a single field or vice versa)

I couldn't tell if we'd need an interface called `Default` for every POJO, or whether this was a one-off. Regardless of this I found it confusing.

Finally we lost a constructor which made the tests (slightly) more complex because we had to specify more nulls.

krdabrowski:
More complex mappings can be done by simply implementing the logic in custom method, or using `@Mapping` annotation - like in this class: https://github.com/Alfresco/hxinsight-connector/pull/493/files#diff-220ba94b1c877c981782f3052e3b0820b75dbfe14859dc070a3d148399e12832

The `@Default` annotation is required in this case, because we have a POJO without setters and with multiple constructors (`NoArgs` and `AllArgsa`). I'm telling MapStruct this way which constructor should be picked up. By default MapStruct picks up (if it's present) default (no-arg) constructor and uses setters.

Regarding lost constructor it's NOT related to MapStruct - I've decided that it's not necessary to maintain an extra constructor only for tests, but I can revert this change.

krdabrowski:
Code adjusted according to your recommendation

tpage-alfresco (approved):
LGTM

mpichura (approved):
I am in favour of introducing `mapstruct` as I find it almost equally as useful as `lombok`.
It's nice that it also works fine with JAMP as provided dependency as it creates implementation classes which get packaged.
So if you ask me - I am for it.

Pawel-608 (approved):
LGTM

PR #497, opened 2024-07-18
ACS-8163: add ACL properties to nodes in bulk ingester

Pawel-608:


bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoNodeMapper.java
```diff
@@ -80,6 +88,9 @@ public IngestEvent map(AlfrescoNode alfrescoNode)
         }
         allProperties.put(CREATED_AT_PROPERTY, createdAt);
 
+        allProperties.put(ALLOW_ACCESS, getResourceReaderAuthorities(alfrescoNode));
+        allProperties.put(DENY_ACCESS, getResourceDeniedAuthorities(alfrescoNode));
```
tpage-alfresco, line 92: I think it would be better to omit these fields for community installations. In that way it is really obvious if the document:
* Has ACLs set
* Has no ACLs set
* Does not specify whether there are ACLs set or not

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoNodeMapperTest.java
```diff
@@ -91,7 +102,9 @@ void shouldMapNodeWithoutContentAndProperties()
                 CREATED_BY_PROPERTY, CREATOR_ID,
                 MODIFIED_BY_PROPERTY, MODIFIER_ID,
                 ASPECT_NAMES_PROPERTY, Set.of(PREFIXED_ASPECT_TITLED),
-                CREATED_AT_PROPERTY, CREATED_AT_TIMESTAMP), ingestEvent.properties());
+                CREATED_AT_PROPERTY, CREATED_AT_TIMESTAMP,
+                ALLOW_ACCESS, Set.of(GROUP_EVERYONE),
+                DENY_ACCESS, Set.of(BOB)), ingestEvent.properties());
```
tpage-alfresco, line 107: A bit harsh on Bob maybe?

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoNodeMapper.java
tpage-alfresco: Ah - I just realised that this is for the Bulk Ingester, which uses the ES library and so always includes ACL information. In that case let's leave this as is.

tpage-alfresco (approved):
LGTM

bulk-ingester/src/test/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoNodeMapperTest.java
Pawel-608: 😂

azakrzewski-hy (approved):
LGTMT

mpichura (approved):
LGTM

PR #501, opened 2024-07-19
ACS-8164: add ACL properties to nodes in live ingester

Pawel-608:
Due to our event system limitations ACL list has to be included in every event that we are sending to hxi (as "updated") because we don't have any way to tell if it was really updated or not


Pawel-608:
wrong jira ticket - should be done under https://hyland.atlassian.net/browse/ACS-8175

PR #506, opened 2024-07-19
ACS-8457 Update JAMP to use Java 11.

tpage-alfresco:


azakrzewski-hy (approved):
LGTM

aonych (approved):
LGTM!

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/config/QuestionsApiConfig.java
```diff
@@ -28,11 +28,21 @@
...
-        int maxContextSizeForQuestion)
+import lombok.Data;
+
+@Data
+public class QuestionsApiConfig
 {
-    public QuestionsApiConfig
+    private int maxContextSizeForQuestion;
+
+    public QuestionsApiConfig(int maxContextSizeForQuestion)
```
mpichura, line 38: I assume this constructor (and the setter below too) is "stronger" than lombok's `@Data`?

mpichura (approved):
LGTM, one doesn't even realize how many post Java 11 feature we use...

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/rest/api/config/QuestionsApiConfig.java
tpage-alfresco: Lombok generates a `RequiredArgsConstructor` which in this case should be a default constructor. I overrode the setter since I wanted to preserve the validation.

PR #510, opened 2024-07-22
ACS-8460: Create test that JAMP works with Java 11

krdabrowski:
https://hyland.atlassian.net/browse/ACS-8460


pom.xml
```diff
@@ -51,7 +51,12 @@
     </distributionManagement>
 
     <properties>
-        <maven.build.sourceVersion>17</maven.build.sourceVersion>
+        <java.version>17</java.version>
+        <repo.java.version>11</repo.java.version>
+        <test.java.version>${java.version}</test.java.version>
```
tpage-alfresco, line 56: I don't think this is used (except three lines below). Can we remove it?

tpage-alfresco:
This looks like you're testing the right thing, but I'm not sure it's particularly valuable. I think it's more pressing that we do something about our build time. To build d0caed8 took 5:19:46 of build time! Most of that is duplicated because we're testing the branch and some arbitrary merge commit that GitHub invented.

This change added about an hour and a half to the build agent usage!

.github/workflows/ci.yml
```diff
@@ -184,7 +184,7 @@ jobs:
           password: ${{ secrets.QUAY_PASSWORD }}
       - uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v5.33.0
       - name: "Build application"
-        run: mvn ${{ env.MAVEN_CLI_OPTS }} clean install -DskipTests -Dalfresco-platform.version=${{ matrix.repoVersion }}
+        run: mvn ${{ env.MAVEN_CLI_OPTS }} clean install -DskipTests -Dalfresco-platform.version=${{ matrix.repoVersion }} -Drepo.java.version=17
```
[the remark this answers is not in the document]
krdabrowski: 🤔 You are right - this is not necessary here - fixed

krdabrowski:
I've restricted the test trigger to `push` events and `master` branch - this should help with the build time.

Pawel-608 (approved):
LGTM

mpichura (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -204,6 +205,55 @@ jobs:
...
+      - uses: Alfresco/alfresco-build-tools/.github/actions/free-hosted-runner-disk-space@v5.33.0
+      - name: "Login to Quay.io"
+        uses: docker/login-action@v3.2.0
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+      - uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v5.33.0
+      - name: "Build application"
+        if: env.JAVA_VERSION_SUPPORTED != env.JAVA_VERSION
```
Pawel-608, line 234: maybe JAVA_11? and btw - will it always be false?

Pawel-608: I don't get the idea of these checks😂

Pawel-608 (commented):
hmm, imo it would be simpler to just hard-code target version to java 11 where needed and then just test if we can build the app with both java 11 and 17🤔 but maybe I'm not getting something

.github/workflows/ci.yml
krdabrowski: It depends if in future we will have the need to again support some older version of Java then used in project 🤔

krdabrowski: After our support for ACS 7.4 will finish we can then fix `JAVA_VERSION_SUPPORTED=17`, but maybe this job could be then removed 🤔 and this ifs won't be necessary 🤔

krdabrowski: I wanted to came up with kind of a universal solution for CI, and disable whole job if connector and alfresco java version were the same, but GHA `job.if` and `job.name` statements doesn't have access to `env` context 🙄 What is funny `job.steps.if` statement has access to `env`

Pawel-608: it sounds unlikely that we'll need to add support for other java versions🤔 @mpichura wdyt? I would keep it as simple as possible and just hardcode these versions

krdabrowski: Thanks to @damianujma hint I was able to fully parametrize this job, so it behaves as expected - run only for multiple supported Java versions; in different case don't even start

PR #520, opened 2024-07-25
ACS-8464 moving user agent header logic to common-authentication

mpichura:
First part - just moving most of the user-agent logic to a common module. Next parts will come with adding user agent headers in prediction applier and JAMP.
Let me know if we need any new "common" module for this or I can move it to `common-authentication` module (or you maybe have any ideas on how to rename `common-authentication` module?).


common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/auth/AuthService.java
```diff
@@ -51,35 +51,43 @@ public void setAlfrescoAuthorizationHeaders(Exchange exchange)
...
     }
 
-    private String getAlfrescoAuthHeaderValue(AccessTokenProvider accessTokenProvider, AuthProperties.AuthProvider authProvider)
+    public String getAuthHeaderValue(String providerId)
+    {
+        AuthProperties.AuthProvider authProvider = authProperties.getProviders().get(ALFRESCO_AUTH_PROVIDER);
+        return providerId.equals(HXI_AUTH_PROVIDER) ? getBearerAuthHeader(HXI_AUTH_PROVIDER) : getAlfrescoAuthHeaderValue(authProvider);
+    }
+
+    private String getAlfrescoAuthHeaderValue(AuthProperties.AuthProvider authProvider)
```
krdabrowski, line 74: This `Value` suffix is required here? Other method: `getBearerAuthHeader()` doesn't have it in it's name 🤔

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java
```diff
@@ -131,6 +132,14 @@ public static void beforeAll()
         WireMock.configureFor(hxInsightMock.getHost(), hxInsightMock.getPort());
     }
 
+    @AfterEach
+    public void reset()
```
krdabrowski, line 136: Maybe method name could be more descriptive like: `resetWiremock()`?

krdabrowski (approved):
LGTM, I only have small remarks

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/auth/AuthService.java
mpichura: I don't think I changed the name of that method with this PR, but it's a good point.
On the other hand, none of these `get*Header*` methods is returns headers per se but only their values.
I am happy to rename all of them to be more consistent (and remove `value` part from all of them if that's ok).

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java
mpichura: I'm not sure if that method is even needed - I've added it recently as I thought it'd fix the failing E2E test - but it doesn't seem to help. What is worse - on my local machine, I can't get the same test to fail...

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/auth/AuthService.java
tpage-alfresco: Fine with me too :+1:

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/api/DiscoveryApiClient.java
```diff
@@ -0,0 +1,86 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import org.alfresco.hxi_connector.common.adapters.auth.AuthService;
+import org.alfresco.hxi_connector.common.util.ErrorUtils;
+
+@RequiredArgsConstructor
+@Slf4j
+public class DiscoveryApiClient
+{
+    static final int EXPECTED_STATUS_CODE = 200;
```
tpage-alfresco, line 49: I've been using the Apache `SC_OK` constants for this.

tpage-alfresco (approved):
LGTM

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/api/DiscoveryApiResponse.java
```diff
@@ -0,0 +1,87 @@
...
+@NoArgsConstructor
+public class DiscoveryApiResponse
+{
+    @NotBlank
+    private RepositoryInfoEntry entry;
+
+    public String getFullVersion()
+    {
+        return entry.getRepository().getVersion().getMajor() + "." +
+                entry.getRepository().getVersion().getMinor() + "." + entry.getRepository().getVersion().getPatch();
```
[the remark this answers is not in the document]
mpichura: Version parts have not been changed in Discovery API when we changed versioning schema to `YY.SP`.
I checked that with several `7.x` and `23.x` versions (as I was investigating why ACS `23.2.1` returns `23.2.0` as version)

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/api/DiscoveryApiClient.java
mpichura: Good point - will do.

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java
mpichura: I think I've found a descent way to fix that failing test and will use that method, so I'll rename it.

PR #526, opened 2024-07-26
ACS-8351 Add implementation to get Agent Avatar

aonych:
Added implementation to previously created API for agent avatar. To use `BinaryResource` we must first create a temporary file.

**Future Improvement Idea**: Adding a new class to the `alfresco rest api` to enable `BinaryResource` to be created from a byte array.


hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
```diff
@@ -168,4 +171,28 @@ public String retryQuestion(String questionId, String comments, Question questio
...
+                    .headers(authService.getAuthHeaders())
+                    .GET()
+                    .build();
+
+            HttpResponse<InputStream> httpResponse = client.send(request, BodyHandlers.ofInputStream());
+
+            ensureCorrectHttpStatusReturned(SC_OK, httpResponse);
+            log.atDebug().log("Agent with id {} received avatar successfully", agentId);
+
+            File tempImageFile = TempFileProvider.createTempFile(httpResponse.body(), "RenditionsApi-", "png");
```
damianujma, line 190: Since the Agent Avatar is not related to `RenditionsApi` we should change the name prefix of the temporary file (e.g. we can use UUID).

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
```diff
@@ -168,4 +171,28 @@ public String retryQuestion(String questionId, String comments, Question questio
...
+            return new FileBinaryResource(tempImageFile, null);
```
damianujma, line 191: We can use a constructor with a single argument.

damianujma (changes requested):

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
aonych: Thank you! Done.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
aonych: Done :)

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/AgentsE2eTest.java
```diff
@@ -91,6 +100,22 @@ void shouldReturnAgents()
         assertEquals(expected, response.jsonPath().get("list.entries"));
     }
 
+    @Test
+    void shouldReturnAgentAvatar() throws IOException
```
dsibilio, line 104: Shouldn't we also cover more scenarios via E2E tests? 

E.g.:
- trying to request an avatar other than `-default-` _(that's already unit-tested, but I'd vote in favor of having the E2E test verify it as well so it can also verify the **response status code**!)_
- trying to request an avatar for an agentId that doesn't exist

damianujma (approved):
LGTM.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
```diff
@@ -26,14 +26,16 @@
...
+
 import static org.apache.http.HttpStatus.SC_ACCEPTED;
 import static org.apache.http.HttpStatus.SC_OK;
 import static org.apache.http.HttpStatus.SC_SERVICE_UNAVAILABLE;
 
 import static org.alfresco.hxi_connector.hxi_extension.service.model.FeedbackType.RETRY;
 import static org.alfresco.hxi_connector.hxi_extension.service.util.HttpUtils.ensureCorrectHttpStatusReturned;
 
-import java.io.IOException;
+import java.io.*;
```
dsibilio, line 38: It'd be neater to import exactly what we need rather than importing the whole package.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
```diff
@@ -168,4 +173,28 @@ public String retryQuestion(String questionId, String comments, Question questio
...
+
+            ensureCorrectHttpStatusReturned(SC_OK, httpResponse);
+            log.atDebug().log("Agent with id {} received avatar successfully", agentId);
+
+            File tempImageFile = TempFileProvider.createTempFile(httpResponse.body(), format("avatar-%s", agentId), "png");
+            return new FileBinaryResource(tempImageFile);
+        }
+        catch (Exception e)
+        {
+            throw new WebScriptException(SC_SERVICE_UNAVAILABLE, format("Failed to get avatar for agent with id %s", agentId), e);
```
dsibilio, line 197: When we throw this exception, how does the response look like to the client that requested an avatar with e.g.: a bad agent Id?

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
```diff
@@ -168,4 +173,28 @@ public String retryQuestion(String questionId, String comments, Question questio
...
+            HttpRequest request = HttpRequest.newBuilder()
+                    .uri(URI.create(format(config.getAvatarUrl(), agentId)))
+                    .headers(authService.getAuthHeaders())
+                    .GET()
+                    .build();
+
+            HttpResponse<InputStream> httpResponse = client.send(request, BodyHandlers.ofInputStream());
+
+            ensureCorrectHttpStatusReturned(SC_OK, httpResponse);
+            log.atDebug().log("Agent with id {} received avatar successfully", agentId);
```
dsibilio, line 190:
```suggestion
            log.atDebug().log("Successfully retrieved avatar for Agent with id: {}", agentId);
```

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
```diff
@@ -168,4 +173,28 @@ public String retryQuestion(String questionId, String comments, Question questio
...
+
+            File tempImageFile = TempFileProvider.createTempFile(httpResponse.body(), format("avatar-%s", agentId), "png");
```
dsibilio, line 192: What happens here if we request multiple times the avatar for the same agentId? Does the `TempFileProvider` implementation include randomness? Will we have a conflict on trying to store the same file twice? Does the implementation override the previous file or leave the previous one, if no randomness is there?

dsibilio (changes requested):

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
dsibilio: **Note:** should be verified via E2E test if possible IMO, as requested here: https://github.com/Alfresco/hxinsight-connector/pull/526#discussion_r1694777577

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
mpichura: I'm surprised PMD scan didn't catch that. I thought we had a rule for that.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
```diff
@@ -168,4 +173,28 @@ public String retryQuestion(String questionId, String comments, Question questio
                 .build());
         return askQuestion(question);
     }
+
+    public BinaryResource getAvatar(String agentId)
```
mpichura, line 177: It would be worth caching this methods results as they are quite static and shouldn't change much. 
That doesn't have to come with this PR as I think getting avatars won't be very frequently called but still might be worth it.

mpichura, line 177: It would be worth caching this methods results as they are quite static and shouldn't change much. 
That doesn't have to come with this PR as I think getting avatars won't be very frequently called but still might be worth it.
Caching might also come handy when we want to reduce temporary file creation needs.

mpichura (commented):
I had a few remarks and agree with Domenico that we could use some more E2E tests cases covered.

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/AgentsE2eTest.java
aonych: Added 👍

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
aonych: Added e2e test to cover it.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
aonych: I changed the file prefix to be unique.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
aonych: I think it's a good idea. I'll prepare a separate PR with this.

mikolajbrzezinski (approved):

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/AgentsE2eTest.java
```diff
@@ -91,6 +104,56 @@ void shouldReturnAgents()
...
+                .then().extract().response();
+
+        // then
+        assertEquals(SC_OK, response.statusCode());
+        assertArrayEquals(Files.readAllBytes(Paths.get("src/test/resources/wiremock/hxinsight/__files/avatar.png")),
+                IOUtils.toByteArray(response.body().asInputStream()));
+    }
+
+    @Test
+    void shouldReturn404ForAvatarWithNotDefaultId() throws IOException
```
dsibilio, line 124:
```suggestion
    void shouldReturn404ForAvatarWithoutDefaultId() throws IOException
```

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/AgentsE2eTest.java
```diff
@@ -91,6 +104,56 @@ void shouldReturnAgents()
...
+                .when().get(repository.getBaseUrl() + format("/alfresco/api/-default-/private/hxi/versions/1/agents/%s/avatars/%s", AGENT_ID, avatarId))
+                .then().extract().response();
+
+        // then
+        assertEquals(SC_NOT_FOUND, response.statusCode());
+        assertTrue(((String) response.jsonPath().get("error.briefSummary")).contains(format("Avatar with id=%s not found", avatarId)));
+    }
+
+    @Test
+    void shouldNotGetAvatarForNotExistingAgent() throws IOException
```
dsibilio, line 141:
```suggestion
    void shouldNotGetAvatarForNonExistentAgent() throws IOException
```

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
```diff
@@ -168,4 +175,29 @@ public String retryQuestion(String questionId, String comments, Question questio
...
+                    .headers(authService.getAuthHeaders())
+                    .GET()
+                    .build();
+
+            HttpResponse<InputStream> httpResponse = client.send(request, BodyHandlers.ofInputStream());
+
+            ensureCorrectHttpStatusReturned(SC_OK, httpResponse);
+            log.atDebug().log("Successfully retrieved avatar for Agent with id: {}", agentId);
+
+            String filePrefix = format("avatar-%s-%s", agentId, System.currentTimeMillis());
```
dsibilio, line 194: Let's use a UUID to add randomness instead. It'd be safer than `currentTimeMillis`.

dsibilio (changes requested):

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
aonych: Thank you, changed.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
```diff
@@ -191,8 +192,7 @@ public BinaryResource getAvatar(String agentId)
             ensureCorrectHttpStatusReturned(SC_OK, httpResponse);
             log.atDebug().log("Successfully retrieved avatar for Agent with id: {}", agentId);
 
-            String filePrefix = format("avatar-%s-%s", agentId, System.currentTimeMillis());
-            File tempImageFile = TempFileProvider.createTempFile(httpResponse.body(), filePrefix, "png");
+            File tempImageFile = TempFileProvider.createTempFile(httpResponse.body(), UUID.randomUUID().toString(), "png");
```
dsibilio, line 195: It would still be ideal to add a prefix to the uuid so that these temp files can easily be identified e.g.: via `ls avatar-*`

aonych: That's right, I accidentally deleted that. Thanks!

PR #531, opened 2024-07-29
ACS-8464 user agent header in prediction applier

mpichura:


damianujma (approved):

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/ApplicationInfoProvider.java
```diff
@@ -52,8 +52,10 @@ public String getUserAgentData()
     {
         if (StringUtils.isBlank(applicationInfo))
         {
+            log.debug("Calculating user agent data.");
```
krdabrowski, line 55:
```suggestion
            log.atDebug().log("Calculating user agent data.");
```

prediction-applier/src/integration-test/java/org/alfresco/hxi_connector/prediction_applier/domain/usecase/e2e/PredictionApplicationIntegrationTest.java
```diff
@@ -25,30 +25,57 @@
...
     {
         // given
         containerSupport.prepareHxInsightToReturnPredictionBatch(BATCH_ID, NODE_ID, PREDICTED_VALUE);
+        containerSupport.prepareRepositoryToReturnDiscovery();
 
         // when
         triggerPredictionsCollection();
 
         // then
+        containerSupport.expectGetBatchesCalled();
```
krdabrowski, line 59: Nice catch!

krdabrowski (approved):
LGTM

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/ApplicationInfoProvider.java
```diff
@@ -52,8 +52,10 @@ public String getUserAgentData()
...
             applicationInfo = calculateUserAgentData();
         }
+        log.atDebug().log("User agent data: {}", applicationInfo);
```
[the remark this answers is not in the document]
dsibilio: To be precise both usages of `log.debug` / `log.atDebug` in this PR could have been `log.debug` instead because there is no additional processing done in these cases.

`log.debug("Calculating user agent data.");` is just logging a static string
`log.atDebug().log("User agent data: {}", applicationInfo);` is just logging a string where `applicationInfo` has already been fully resolved via `calculateUserAgentData` and no further method invocations are being done on it _(e.g.: `applicationInfo.getSomething()`)_

Even then, I very much doubt all these concerns about performance are of any real significance, so we can choose to proceed as we see fit. If we think it's easier to just spam `log.atDebug().log(...)` even though it's unnecessary, then we can do it.

prediction-applier/src/integration-test/java/org/alfresco/hxi_connector/prediction_applier/domain/usecase/e2e/PredictionApplicationIntegrationTest.java
```diff
@@ -25,30 +25,57 @@
...
 package org.alfresco.hxi_connector.prediction_applier.domain.usecase.e2e;
 
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
+import org.springframework.test.annotation.DirtiesContext;
 
 import org.alfresco.hxi_connector.prediction_applier.domain.usecase.e2e.util.PredictionApplierE2ETestBase;
 
 @SuppressWarnings("PMD.JUnitTestsShouldIncludeAssert")
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
```
dsibilio, line 35: Isn't there any way to avoid this? E.g.: manually resetting some dirtied properties / using beforeAll | afterAll appropriately? I'm not entirely aware of the reason why we added this, but it can be quite expensive as more methods get added to this class.

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/config/AppConfig.java
```diff
@@ -0,0 +1,62 @@
...
+    public Application application()
+    {
+        return new Application();
+    }
+
+    @Bean
+    public DiscoveryApiClient discoveryApiClient(RepositoryApiProperties repositoryApiProperties, AuthService authService, ObjectMapper objectMapper)
+    {
+        return new DiscoveryApiClient(repositoryApiProperties.discoveryEndpoint(), authService, objectMapper);
+    }
```
dsibilio, line 55: I seem to remember the `DiscoveryApiClient` had been declared as an optional bean conditional on the presence of the Repository URL.

Is there no way to replicate / respect that logic here as well, so that we can alternatively rely on the version override if needed?

dsibilio: cc: @damianujma

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/config/AppConfig.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Bean
+    public ApplicationInfoProvider applicationInfoProvider(DiscoveryApiClient discoveryApiClient, Application applicationProperties)
+    {
+        return new ApplicationInfoProvider(discoveryApiClient, applicationProperties, Optional.empty());
```
dsibilio, line 60: Same here, the `versionOverride` doesn't seem to be taken into consideration 🤔

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/hx_insight/HxInsightUrlProducer.java
```diff
@@ -46,16 +47,21 @@ public class HxInsightUrlProducer
...
+        return withUserAgentDataPathParam(PREDICTIONS_URL_PATTERN.formatted(insightPredictionsProperties.sourceBaseUrl(), BATCH_ID_HEADER, PREDICTIONS_PAGE_NO_HEADER));
     }
 
     public String getConfirmationUrl()
     {
-        return PREDICTIONS_CONFIRMATION_URL_PATTERN.formatted(insightPredictionsProperties.sourceBaseUrl(), BATCH_ID_HEADER);
+        return withUserAgentDataPathParam(PREDICTIONS_CONFIRMATION_URL_PATTERN.formatted(insightPredictionsProperties.sourceBaseUrl(), BATCH_ID_HEADER));
+    }
+
+    private String withUserAgentDataPathParam(String url)
```
dsibilio, line 63:
```suggestion
    private static String withUserAgentDataPathParam(String url)
```

prediction-applier/src/integration-test/java/org/alfresco/hxi_connector/prediction_applier/domain/usecase/e2e/PredictionApplicationIntegrationTest.java
mpichura: At the moment - there isn't. User agent data is lazy loaded and is not being refreshed ever since as application is running (maybe we could do that from time to time, but still, I treat it as kind of static data)
If the test method which executes a happy path is run before the second method, then application context already has repository version pulled from ACS and will not fail, however we'd like it to fail on that in prediction applier.

prediction-applier/src/integration-test/java/org/alfresco/hxi_connector/prediction_applier/util/ContainerSupport.java
```diff
@@ -202,6 +208,69 @@ public void expectBatchStatusWasUpdated(String batchId, String status, int page)
...
+    }
+
+    @SneakyThrows
+    public void expectNoPredictionBatchesRequestsReceived()
+    {
+        TimeUnit.MILLISECONDS.sleep(200);
+        hxInsightMock.verifyThat(exactly(0), postRequestedFor(urlPathEqualTo(HXI_PREDICTION_BATCHES_ENDPOINT)));
+        hxInsightMock.verifyThat(exactly(0), getRequestedFor(urlPathEqualTo(HXI_PREDICTION_BATCHES_ENDPOINT)));
+        resetWireMock();
+    }
```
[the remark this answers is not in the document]
mpichura: Yes, I agree, I also see a little value in abstracting some part of this when the abstraction has only this single child.

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/ApplicationInfoProvider.java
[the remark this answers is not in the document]
mpichura: Domenico is right about keeping `log.debug` as sufficient enough here - I can change that.
And I also doubt there is any significant performance implication.
BTW, I think PMD complained about not using `log.atDebug()` in cases where there actually is some additional processing - we should probably keep that usage to minimal, WDYT?

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/config/AppConfig.java
mpichura: AFAIK, in `prediction-applier` module, we will not use application property to let admins override ACS version, since this module needs connection to running ACS instance.

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/config/AppConfig.java
mpichura: See my previous comment.
BTW, in the next PR (adding user agent header to `hxinsight-extension` module) I am preparing a small refactoring with an abstraction layer put over getting repo version as all of the modules have different approaches/needs to getting the repo version.
I plan to create different beans conditionally on properties in `live-ingester` module, a bean basing on ACS Discovery API in `prediction-applier` module and a bean in `hxinsight-extension` module using a Java API.

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/config/AppConfig.java
dsibilio: Fair enough, thanks for explaining!

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/ApplicationInfoProvider.java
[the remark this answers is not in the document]
dsibilio: I'd be ok with only using the more verbose `log.atDebug().log(...)` syntax only when actually useful.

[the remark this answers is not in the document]
dsibilio: _(PMD should spot that)_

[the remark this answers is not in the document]
mpichura: Fixed  - only `log.debug` is used now.

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/ApplicationInfoProvider.java
mpichura: I went with `log.debug` approach where `log.atDebug()` is not necessary.

PR #534, opened 2024-07-30
NO_JIRA: do not grow test delay

Pawel-608:
imo this is better approach because:
`+` you can predict how much time test needs pretty accurately
`+` it does not slow down tests that much if they are failing

`-` might need some adjustments of delay in some test


mpichura (approved):
LGTM

PR #546, opened 2024-08-01
ACS-8505: Revert docker-compose run for CreateNodeE2eTest

krdabrowski:
https://hyland.atlassian.net/browse/ACS-8505


prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/repository/NodesClient.java
```diff
@@ -110,6 +111,7 @@ public void configure()
             .when(header(HTTP_RESPONSE_CODE).isNotEqualTo(String.valueOf(EXPECTED_STATUS_CODE)))
                 .process(this::throwExceptionOnUnexpectedStatusCode)
             .otherwise()
+                .log(DEBUG, log, "Node updated successfully - Headers: ${headers}, Body: ${body}")
```
mpichura, line 114: I'm not sure if we should log message headers/body at `debug` level - they may contain sensitive information (I am not sure if we should log that at all, but if so - maybe at `trace`)
Besides that - should we delete/change log message 3 lines lower?

mpichura (approved):
Great job. I had one remark about logging. 
Also - I'm thinking whether we should or not keep `docker-compose-minimal.yml` (I tend to use it from time to time when I don't need full env with UI), WDYT?

prediction-applier/src/main/java/org/alfresco/hxi_connector/prediction_applier/repository/NodesClient.java
krdabrowski: This was temporary solution to investigate logs - it's already fixed.
I can revert the minimal d-c file 🤔

krdabrowski: Minimal d-c file has been restored, but in a minimalist version - with references to main d-c setup

PR #547, opened 2024-08-01
ACS-8464 user agent header in hxinsight JAMP

mpichura:
Also abstracting ACS version retrieval.


hxinsight-extension/src/main/resources/alfresco/module/alfresco-hxinsight-connector-hxinsight-extension/context/service-context.xml
```diff
@@ -174,4 +174,15 @@
     <bean class="org.alfresco.hxi_connector.hxi_extension.rest.api.QuestionFeedbackRelation">
         <constructor-arg name="hxInsightClient" ref="hxInsightClient"/>
     </bean>
+
+    <bean id="acsRepositoryInformation" class="org.alfresco.hxi_connector.hxi_extension.service.AcsRepositoryInformation">
+        <constructor-arg name="descriptorService" ref="DescriptorService"/>
+    </bean>
+    <bean id="hxiApplication" class="org.alfresco.hxi_connector.common.config.properties.Application">
```
krdabrowski, line 181: The bean name (and also property bellow) is confusing for me - I assume it's connector version, not some HxI AI app version

krdabrowski (approved):
LGTM, I have only one remark

damianujma (approved):
LGTM.

hxinsight-extension/src/main/resources/alfresco/module/alfresco-hxinsight-connector-hxinsight-extension/context/service-context.xml
mpichura: Changed - hopefully new names are better suiting and less confusing.

aonych (approved):

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
```diff
@@ -104,7 +109,8 @@ public String askQuestion(Question question)
 
             HttpRequest request = HttpRequest.newBuilder()
                     .uri(URI.create(config.getQuestionUrl()))
-                    .header("Content-Type", "application/json")
+                    .header(CONTENT_TYPE, "application/json")
+                    .header(USER_AGENT, applicationInfoProvider.getUserAgentData())
```
dsibilio, line 113: I wonder if we could have solved this with a solution that is more fit to address cross-cutting concerns such as this one, e.g.: using an HttpRequestInterceptor?

I'm personally ok with this solution, but it is error-prone in the sense that all new APIs that are implemented rely on the developers to remember they have to manually add this user agent calculation into the request.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/messaging/LiveIngesterMessagingConfig.java
```diff
@@ -62,20 +63,24 @@ public Application application()
...
+        else if (StringUtils.isNotBlank(integrationProperties.alfresco().repository().discoveryEndpoint()))
+        {
+            return new DiscoveryApiRepositoryInformation(integrationProperties.alfresco().repository().discoveryEndpoint(), authService, objectMapper, HttpClient.newHttpClient());
+        }
+        throw new IllegalStateException("Either property alfresco.repository.discovery-endpoint or alfresco.repository.version-override must be set in the Live Ingester configuration.");
     }
 
     @Bean
-    public ApplicationInfoProvider applicationInfoProvider(DiscoveryApiClient discoveryApiClient, IntegrationProperties integrationProperties)
+    @DependsOn("repositoryInformation")
```
dsibilio, line 80: I find it odd that this `DependsOn` was necessary even though `RepositoryInformation` is one of the required arguments (read: beans) of the method. Could this be an intermediate change that is not required anymore?

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/DiscoveryApiRepositoryInformation.java
```diff
@@ -81,7 +85,7 @@ private DiscoveryApiResponse getDiscoverApiResponse() throws IOException, Interr
         throwExceptionOnUnexpectedStatusCode(response.statusCode(), SC_OK);
 
         DiscoveryApiResponse discoveryApiResponse = objectMapper.readValue(response.body(), DiscoveryApiResponse.class);
-        log.atTrace().log("Discovery API response: {}", discoveryApiResponse);
+        log.trace("Discovery API response: {}", discoveryApiResponse);
```
dsibilio, line 88: I meant to ask, is this API only invoked once at the startup to determine the version, or is it invoked N times? If it's the latter, then we should - at least eventually - plan caching this IMO.

dsibilio (commented):
Mostly LGTM, left a few comments with doubts/requests for clarifications and suggestions for potential improvements. I'll leave it up to you to decide whether it's appropriate to address those now or it would be best to track them and deal with them later on.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
mpichura: I tried to use a more reliable solution but with `java.net.http.HttpClient` there's not much to do here.
I guess the best thing we could possibly do is some custom implementation (either for `java.net.http.HttpRequest` or maybe `java.net.http.HttpClient`) which would cover that. 
I'll push some more generic approach for this - let me know what you think.

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/DiscoveryApiRepositoryInformation.java
mpichura: Thanks for catching the logger part.
About caching - `ApplicationInfoProvider` class which is used across all modules as `User-Agent` header value provider will call `RepositoryInformation::getRepositoryVersion` only once and keep it in class member field.
If this solution is somehow faulty, we can plan some caching, but for now, I don't see a reason to do so.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/messaging/LiveIngesterMessagingConfig.java
mpichura: Thanks for catching that - probably a leftover from some try-and-test - removing that annotation.

hxinsight-extension/src/main/java/org/alfresco/hxi_connector/hxi_extension/service/HxInsightClient.java
dsibilio: We could extend the `HttpClient#send` method to automatically decorate all requests, creating an e.g.: `HxHttpClient` class.

Similar to https://stackoverflow.com/questions/53215038/how-to-log-request-response-using-java-net-http-httpclient/55455589#55455589.

PR #553, opened 2024-08-02
ACS-8179: test live permission updates

Pawel-608:


e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeLiveIngesterE2eTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @AfterEach
+    void tearDown()
+    {
+        WireMock.reset();
+    }
+
+    @Test
+    @SuppressWarnings({"PMD.JUnitTestsShouldIncludeAssert"})
+    final void testCreateNodeWithDefaultPermissions()
```
krdabrowski, line 124: I'm not getting why for this test was created separate class - it couldn't be placed in [UpdateNodeE2eTest](https://github.com/Alfresco/hxinsight-connector/blob/4a19a9bbec1743f475d9e7016c4f2a043bd39c3b/e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java)?

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/util/TestJsonUtils.java
```diff
@@ -0,0 +1,49 @@
...
+import lombok.SneakyThrows;
+
+@NoArgsConstructor(access = PRIVATE)
+@SuppressWarnings({"PMD.TestClassWithoutTestCases", "PMD.FieldNamingConventions"})
+public final class TestJsonUtils
+{
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    @SneakyThrows
+    public static Set<String> getSetProperty(JsonNode jsonNode, String propertyName)
```
krdabrowski, line 45: Method name is really confusing 😅 What this method does? 🧐

Pawel-608: it returns property of type set😂 maybe I should just create "toSet(String)" method instead?

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeLiveIngesterE2eTest.java
Pawel-608: I would say that we should split e2e tests because in fact we have 3 separate flows:

bulk ingester -> live ingester -> hxi

repo -> live ingester -> hxi

hxi -> prediction applier -> repo

and imo mixing them is not a good approach (eventually we could mix first two but definitely not 3-rd one)

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/util/TestJsonUtils.java
krdabrowski: This sounds better, I'm for 🙋‍♂️

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeLiveIngesterE2eTest.java
krdabrowski: For component "endpoint2enpoint" testing we have those: https://github.com/Alfresco/hxinsight-connector/tree/ff1d1f9ba5391db64ae854153643d726a24bdaf3/live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository

krdabrowski (approved):
LGTM

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeLiveIngesterE2eTest.java
Pawel-608: moved!

PR #557, opened 2024-08-05
ACS-8370 Set temporary outdated image

damianujma:


dsibilio (approved):
Let's merge for testing, trigger DependaBot and then see how that behaves 🤞

PR #570, opened 2024-08-08
NO_JIRA: enable tests in PRs created by dependabot

krdabrowski:


mpichura (approved):

.github/workflows/ci.yml
```diff
@@ -146,7 +146,7 @@ jobs:
     needs:
       - pre_commit
     if: >
-      github.event_name == 'push' &&
+      (github.event_name == 'push' || github.actor == 'dependabot[bot]') &&
```
dsibilio, line 149: Maybe we should simply reconsider having `github.event_name == 'push'` in the first place.

dsibilio (changes requested):
Left a suggestion to remove the push-based logic, rather than adding further conditionals for DependaBot PRs. Unless there was a good reason why these tests shouldn't be run in a PR build!

.github/workflows/ci.yml
krdabrowski: I'm also in favor of running the PR builds. I think that Tom's intention was to limit the build time by disabling duplicated tests, but IMO PR build are more valuable then push ones. We should switch to PR builds, or at least re-enable them 🤔  WDYT @mpichura ?

mpichura: You know I'm in favour of running heavy tests on PR build rather than push. I just don't want to make such changes while Tom's on PTO.

aonych (approved):

.github/workflows/ci.yml
krdabrowski: Ok, so let merge it as it is and return to this topic after Tom's return from PTO. Do you agree @dsibilio ?

dsibilio: Sure, go ahead

dsibilio (approved):
LGTM although I think it'd make more sense to run these tests always (both push and PR).

PR #574, opened 2024-08-09
Test if e2e tests will fail on Java 11 with ACS 23.2

krdabrowski:


krdabrowski:
Closing as this is only a temporary testing PR

PR #600, opened 2024-08-21
ACS-8135: add e2e-s for sending timestamp to hxi

Pawel-608:


e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java
```diff
@@ -226,6 +226,20 @@ void testApplyPredictionToUpdatedNode()
...
+    {
+        // when
+        Node updatedNode = repositoryClient.updateNodeWithContent(createdNode.id(), UPDATE_NODE_PROPERTIES);
+
+        // then
+        RetryUtils.retryWithBackoff(() -> verify(exactly(1), postRequestedFor(urlEqualTo("/ingestion-events"))
+                .withRequestBody(containing(updatedNode.id()))
+                .withRequestBody(containing("timestamp"))
+                .withHeader(USER_AGENT, matching(getAppInfoRegex()))));
+    }
```
azakrzewski-hy, line 241: LGTM just small remark: maybe validation that timestamp of updated node is different from timestamp of the original node could be useful

azakrzewski-hy (approved):

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java
Pawel-608: it would be kinda hard to check, but I think that integration and unit tests are covering this part

krdabrowski (approved):

PR #604, opened 2024-08-22
ACS-8607: Create Helm chart for live-ingester - adding liveness probe endpoint

krdabrowski:
https://hyland.atlassian.net/browse/ACS-8607

### DIFF
--- a/live-ingester/src/main/resources/application.yml
+++ b/live-ingester/src/main/resources/application.yml
@@ -2,17 +2,9 @@ server:
   port: 8080
 
 management:
-  endpoints:
-    web:
-      exposure:
-        include: health
-  endpoint:
-    health:
-      probes:
-        enabled: true
-  health:
-    liveness-state:
-      enabled: true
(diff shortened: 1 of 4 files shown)


Pawel-608 (approved):
LGTM

live-ingester/src/main/resources/application.yml
```diff
@@ -1,15 +1,19 @@
-application:
-  source-id: alfresco-dummy-source-id-0a63de491876
-  version: "@project.version@"
+server:
+  port: 8080
```
mpichura, line 2: I'd be surprised if this doesn't result in ports clash when running docker compose from command line.
Can we use a different port by default (or add port mapping to DC files)?

krdabrowski: Live ingester server port is not exposed outside container, so there shouldn't be any issues with port clashes. We expose only debug port

mpichura (approved):
LGTM

PR #605, opened 2024-08-22
NO_JIRA: headers update

krdabrowski:


mpichura (approved):
Let's go with this approach

PR #620, opened 2024-08-27
[ACS-8631] adjust-url-in-wiremock-files

azakrzewski-hy:
[ACS-8631](https://hyland.atlassian.net/browse/ACS-8631)
Adjust url paths in wiremock mapping files for local use


krdabrowski (approved):
LGTM

PR #635, opened 2024-08-30
ACS-8646: add avatar to agents response

Pawel-608:


mpichura (approved):
LGTM.
Although, I'm not sure if we should use real world URLs for avatars (I know these are truly real).

krdabrowski (approved):

PR #641, opened 2024-09-02
ACS-8719: add response completeness field to

Pawel-608:


docs/acs-private-apis.md
```diff
@@ -137,35 +137,24 @@ Get answer for a question.
 * Response:
     * Status: 200
     * Content-Type: `application/json`
-    * Entry object (wrapped with list/pagination information):
```
krdabrowski, line 140: There's only one answer for a question?

krdabrowski: In path `questions/{questionId}/answers` there shouldn't be also present `/-default-` ID?

krdabrowski (approved):
I'm not entirely convinced with this change: list -> entry 🤔 But in the other hand I don't have any serious cons (except required ID in path), so I'm fine with that. WDYT @mpichura ?

docs/acs-private-apis.md
Pawel-608: yep, from what I saw in code there is just one answer but there are multiple "Reference" inside

Pawel-608: fixed path

mpichura (approved):
The change looks okay to me. 
Since HXI returns only one answer, I think we shouldn't wrap it around pagination.

PR #646, opened 2024-09-04
ACS-8732: Adjust docker-compose setup to align with convention

krdabrowski:
Also base keycloak setup on default d-c config


gionn (approved):

mpichura (commented):
This modified DC file with keycloak doesn't seem to work form me.
Both, live-ingester and prediction applier stop at ACS health probe (not sure why, TBH)
When using postman, getting the token from keycloak and then trying to call ACS API  using it, I get a 401 with  
"08050058 Failed to extract username from token: Failed to decode token. An error occurred while attempting to decode the Jwt: The iss claim is not valid. Expected `http://localhost:8080/auth/realms/alfresco` but got `http://localhost:8999/auth/realms/alfresco`.",

distribution/src/main/resources/docker-compose/nginx.conf
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        location /alfresco/ {
+            proxy_pass http://alfresco:8080;
+
+            # If using external proxy / load balancer (for initial redirect if no trailing slash)
+            absolute_redirect off;
+        }
+
+        location /auth/ {
```
gionn, line 72:
```suggestion
        location /auth {
```

@mpichura try with this

trailing slash in location directive usually is not a good thing to have, depending on how you configure proxy_pass (with a trailing slash or not) it may chop or add an additional (double) slash in the final request being forwarded upstream

krdabrowski: Thanks @gionn for pointing this out - I've removed all trailing slashes

mpichura (approved):
Nice job

distribution/src/main/resources/docker-compose/nginx.conf
```diff
@@ -0,0 +1,86 @@
...
+        proxy_set_header Host            $host:$server_port;
+        proxy_set_header X-Real-IP       $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_pass_header Set-Cookie;
+
+        # Protect access to SOLR APIs
+        location ~ ^(/.*/service/api/solr/.*)$ {return 403;}
+        location ~ ^(/.*/s/api/solr/.*)$ {return 403;}
+        location ~ ^(/.*/wcservice/api/solr/.*)$ {return 403;}
+        location ~ ^(/.*/wcs/api/solr/.*)$ {return 403;}
```
tpage-alfresco, line 46: We're not using Solr.

azakrzewski-hy (approved):
LGTM

distribution/src/main/resources/docker-compose/nginx.conf
krdabrowski: Good catch - entries removed

PR #651, opened 2024-09-05
ACS-8729: Increased readability of statusCode verification

krdabrowski:


tpage-alfresco (approved):
LGTM - thanks!

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/AcsHealthProbe.java
```diff
@@ -103,4 +98,9 @@ private void sleep() throws InterruptedException
         log.debug("ACS is not available. Retrying in {} seconds", retryIntervalSeconds);
         TimeUnit.SECONDS.sleep(retryIntervalSeconds);
     }
+
+    private static boolean isErrorCode(int statusCode)
```
mpichura, line 102: Maybe this method could be something like `isNotErrorCode` and we could avoid negating it when it's called.

mpichura (commented):
See my small suggestion

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/AcsHealthProbe.java
krdabrowski: Fixed

PR #666, opened 2024-09-12
ACS-8766: Adjust live-ingester logging to be more descriptive in case of error

krdabrowski:
https://hyland.atlassian.net/browse/ACS-8766


tpage-alfresco (changes requested):
The log quality looks much better :+1: 

I'm not convinced by the method of masking sensitive data out of the log using a regex. Instead I think it would be better to selectively choose data to be included in the log. With the current proposal then there's a risk that the name of a field could change, e.g. `secret` could be renamed `token` by a library upgrade, and then the masking would stop working. I don't think any of the tests here would catch this problem either.

I know it's more work, but I think it would be better to create our own text renderers for the problematic fields.

krdabrowski:
@tpage-alfresco I've adjusted the way of masking stuff according to our conversation on Slack. Please could you review it one more time

mpichura (approved):
LGTM, as discussed, you may want to call `logMaskedExchangeState` direclty in `.process()` instead of creating/calling a private method in each `RouteBuilder` it used.

krdabrowski:
@mpichura fixed - thanks!

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/hx_insight/storage/connector/HttpFileUploader.java
```diff
@@ -112,7 +116,7 @@ STORAGE_LOCATION_HEADER, wrapRawToken(fileUploadRequest.storageLocation()),
                     .withHeaders(headers)
                     .withBody(fileData)
                     .request();
-            log.atDebug().log("Upload :: {} rendition of the node: {} successfully uploaded to pre-signed URL: {}", fileUploadRequest.contentType(), nodeId, fileUploadRequest.storageLocation().getPath());
+            log.info("Storage :: Rendition of type: {} for node: {} successfully uploaded to pre-signed URL: {}", fileUploadRequest.contentType(), nodeId, fileUploadRequest.storageLocation().getPath());
```
tpage-alfresco, line 119: You should use `atInfo()` to avoid the PMD issue (since PMD doesn't know how long the calls to `contentType()`, `storageLocation()` and `getPath()` will take).

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/hx_insight/storage/connector/HttpFileUploader.java
```diff
@@ -123,7 +127,7 @@ STORAGE_LOCATION_HEADER, wrapRawToken(fileUploadRequest.storageLocation()),
             }
             catch (IOException ioe)
             {
-                log.atDebug().log("Upload :: Stream reset failed due to: {}", ioe.getMessage());
+                log.warn("Storage :: Rendition stream NOT reset properly after node %s content upload fail due to: %s".formatted(nodeId, ioe.getMessage()), ioe);
```
tpage-alfresco, line 130: Similarly I think we want `atWarn()` here.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/hx_insight/storage/connector/HttpFileUploader.java
krdabrowski: Fixed - thanks!

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/hx_insight/storage/connector/HttpFileUploader.java
krdabrowski: Fixed - thanks!

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/transform/response/ATSTransformResponseHandler.java
```diff
@@ -50,6 +51,8 @@
 public class ATSTransformResponseHandler extends RouteBuilder
 {
     private static final String ROUTE_ID = "transform-events-consumer";
+    private static final int EXPECTED_STATUS_CODE = 201;
+    private static final String EXPECTED_STATUS_CODE_REGEX = "[\\s\\S]*\"status\"\\s*:\\s*%s[\\s\\S]*".formatted(EXPECTED_STATUS_CODE);
```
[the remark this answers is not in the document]
tpage-alfresco: Explanation: I think that `[\\s\\S]` includes all characters so is the same as `.` Also the status code should not be followed by another digit (and will be followed by something in a JSON response).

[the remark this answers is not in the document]
krdabrowski: The `[\\s\\S]` is similar to `.`, but it also includes whitespaces, so it covers bigger set of characters to match whole sentence / paragraph (or JSON) containing expected status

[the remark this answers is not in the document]
krdabrowski: I can add the `[^0-9]` - this is to prevent the next char from being a number, yes?

tpage-alfresco (approved):
Great - thanks for adding the tests!

PR #696, opened 2024-10-02
[ACS-8876] updating image versions 

azakrzewski-hy:
[ACS-8876](https://hyland.atlassian.net/browse/ACS-8876)
Update image versions for the latest:

ALFRESCO_TAG=23.4.0-M1
SHARE_TAG=23.4.0-M1
TRANSFORM_ROUTER_TAG=4.1.4
TRANSFORM_CORE_AIO_TAG=5.1.4
SHARED_FILE_STORE_TAG=4.1.4
ACTIVE_MQ_TAG=5.18.5-jre17-rockylinux8
DIGITAL_WORKSPACE_TAG=5.1.0
CONTROL_CENTER_TAG=9.1.0
SEARCH_ENTERPRISE_TAG=4.1.0

Update of ADW would allow to add new property:
APP_CONFIG_PLUGIN_KNOWLEDGE_RETRIEVAL: false
for our docker-compose.yml file. This property can be used to enable ability for ADW UI to ask discovery questions.

[ACS-8876]: https://hyland.atlassian.net/browse/ACS-8876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/ci.yml
```diff
@@ -151,7 +151,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        repoVersion: [ 7.3.2, 7.4.2, 23.3.0 ]
+        repoVersion: [ 7.3.2, 7.4.2, 23.3.0, 23.4.0-M1 ]
```
tpage-alfresco, line 154: I don't think we should test both. We're only testing one patch for each LTS version 7.3 and 7.4, and so we should only test one patch for the LTS version 23 (nb. the LTS strategy changed between 7.4.x and 23.x).

azakrzewski-hy: I've rollback the changes for ci.yml file so it keeps 23.3.0

distribution/src/main/resources/docker-compose/docker-compose.yml
```diff
@@ -178,6 +178,7 @@ services:
       APP_CONFIG_PLUGIN_PROCESS_SERVICE: false
       APP_CONFIG_PROVIDER: ECM
       APP_CONFIG_AUTH_TYPE: BASIC
+      APP_CONFIG_PLUGIN_KNOWLEDGE_RETRIEVAL: false
```
krdabrowski, line 181: I think, that by default you can enable KR feature, as this project is all about integration with HxI

azakrzewski-hy: thing is that when project is running with mocked endpoints and we enable KR it will be available on ADW UI but it will not response - BE mocks do not cover FE responses

krdabrowski: Oh! Maybe we should add those mocks? 🤔

azakrzewski-hy: It's a good idea. Maybe we can do that in separate PR

krdabrowski: I'm fine with that

krdabrowski (approved):

PR #763, opened 2024-12-04
[ACS-9062] Switch to ACS Event Model 1.0.2

kmagdziarz:


cezary-witkowski (approved):
Nice job!

damianujma (approved):

krdabrowski:
Seems that some event examples are missing `"@type": "NodeResource"` in `resourceBefore`, eg.
```
"resourceBefore": {
  "name": "Old Folder"
}
```
and `objectMapper` cannot map it and leaves `null` as `resourceBefore` - this is the reason why test a failing.
I'm not sure if `"@type": "NodeResource"` should be now required or not - if yes you need to add it event snippets in `UpdateRequestIntegrationTest`

krdabrowski:
Wired it was working earlier without this `@type` param 😅

PR #799, opened 2025-01-13
PRODSEC-9768: bump tomcat embed version

Pawel-608:


Pawel-608:
we decided not to bump tomcat version for now, see jira for details

PR #805, opened 2025-01-16
[CIN-1052] disable pre_commit for schedule triggered job

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-1052

Disable pre_commit for schedule triggered job as Alfresco/alfresco-build-tools/.github/actions/github-list-changes action does not contain 'schedule' event name - does not supported it


tpage-alfresco (approved):
Thanks Adam

kamilbochno (approved):

PR #810, opened 2025-01-17
feat: run in nightly veracode

PWS-TE:


.github/workflows/ci.yml
```diff
@@ -123,7 +123,7 @@ jobs:
     needs:
       - pre_commit
     if: >
-      (github.event_name == 'push' || github.actor == 'dependabot[bot]') &&
+      (github.event_name == 'push' || github.actor == 'dependabot[bot]' || github.event_name != 'schedule') &&
```
azakrzewski-hy, line 126: Should we have this condition about schedule here?

azakrzewski-hy (commented):
Maybe we should not have negative condition for schedule event in many jobs as all jobs have their own conditions

.github/workflows/ci.yml
```diff
@@ -15,8 +15,6 @@ on:
       - release/**
   schedule:
     - cron: "0 0 * * *"  # Runs every night at midnight
-      branches:
-        - master
```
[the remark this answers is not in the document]
PWS-TE: I agree. I corrected it to PR

.github/workflows/ci.yml
```diff
@@ -35,7 +33,8 @@ env:
 jobs:
   pre_commit:
     runs-on: ubuntu-latest
-    if: github.event_name != 'schedule'
+    if: >
+      (github.ref_name == 'master' || startsWith(github.ref_name, 'release/') || github.event_name == 'pull_request' || github.event_name == 'schedule')
```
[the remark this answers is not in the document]
PWS-TE: I checked and the pipeline also works fine. Added to PR.

.github/workflows/ci.yml
PWS-TE: I can remove them if they are redundant, but I am not sure if these jobs will also start at midnight if there are no conditions in steps

.github/workflows/ci.yml
```diff
@@ -56,7 +54,7 @@ jobs:
     needs:
       - pre_commit
     if: >
-      (github.ref_name == 'master' || startsWith(github.ref_name, 'release/') || github.event_name == 'pull_request') &&
+      (github.ref_name == 'master' || startsWith(github.ref_name, 'release/') || github.event_name == 'schedule') &&
```
tpage-alfresco, line 57: Random extra requirement that I just thought of: It would be useful to continue to run the SCA scan against Dependabot PRs. I guess we'll rarely benefit by running this against our own PRs (we don't often update or include new dependencies), but Dependabot PRs are always doing this.

Even better would be to somehow report on any new issues found as a result of the PR.

tpage-alfresco: I see we have something similar below:
```
... || github.actor == 'dependabot[bot]'
```

.github/workflows/ci.yml
```diff
@@ -73,7 +71,7 @@ jobs:
...
```
tpage-alfresco, line 74: We probably want to continue to run this against PRs. (Although we could exclude Dependabot PRs)

.github/workflows/ci.yml
```diff
@@ -109,6 +107,7 @@ jobs:
     runs-on: ubuntu-latest
     if: >
       github.event_name == 'pull_request' &&
+      github.event_name != 'schedule' &&
```
tpage-alfresco, line 110: Unnecessary. If `event_name == 'pull_request'` then it's not `'schedule'`

.github/workflows/ci.yml
```diff
@@ -35,7 +33,7 @@ env:
 jobs:
   pre_commit:
     runs-on: ubuntu-latest
-    if: github.event_name != 'schedule'
+    if: github.job != 'nightly_tests'
```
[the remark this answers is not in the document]
PWS-TE: We want this step to run with all others except the nightly_tests step. I had to change this because the veracode_sca and veracode_sast steps need the pre_commit step to run and they use schedule. If the condition if: github.event_name != 'schedule' was left then both veracode steps would fail.

.github/workflows/ci.yml
tpage-alfresco: Yep - this is unnecessary too, because dependabot doesn't trigger scheduled builds.

tpage-alfresco (changes requested):
I think we can simplify this PR quite a bit.

.github/workflows/ci.yml
PWS-TE: Thanks, removed unnecessary code.

.github/workflows/ci.yml
PWS-TE: Removed unnecessary code.

.github/workflows/ci.yml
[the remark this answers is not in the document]
PWS-TE: That makes sense. Corrected to github.job = 'pre_commit'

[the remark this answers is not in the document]
azakrzewski-hy: Not sure I understand the logic behind making "if: github.job == 'pre_commit'" condition in "pre_commit" job. If condition inside "pre_commit" job was introduced only because we wanted to skip "pre_commit" job when "nightly_tests" job was triggered. Setting "if: github.job == 'pre_commit'" condition in "pre_commit" job would not allow to skip "pre_commit" job when "nightly_tests" job is running

.github/workflows/ci.yml
PWS-TE: On the line 75 we exclude Dependabot ?

.github/workflows/ci.yml
PWS-TE: Implemented. Please look if it's ok.

.github/workflows/ci.yml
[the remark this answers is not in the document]
PWS-TE: Pre_commit is not necessary to run steps from veracode, so I removed it there. I also split the pipeline into two (night pipeline and day pipeline) to make it more readable.

.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -0,0 +1,96 @@
...
+on:
+  pull_request:
+    branches:
+      - master
+      - release/**
+
+  push:
+    branches:
+      - master
+      - release/**
```
azakrzewski-hy, line 12: Veracode and Nightly Tests are expected to run once a day on scheduled time, can't see any reason to keep "pull_request" and "push" events in this file

azakrzewski-hy (commented):
Splitting jobs into separate files can be a good idea. 
Looking at results of the run after latest changes were pushed into this PR can't see results of ci.yml file tests jobs that we expect to be run

.github/workflows/nightly_tests_and_veracode.yml
PWS-TE: We don't want to check by veracode  master and release branch after merge ?

.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -0,0 +1,87 @@
+name: Veracode and Nightly Tests run
+
+on:
+  schedule:
+    - cron: "0 0 * * *"  # Runs every night at midnight
+
+env:
+  DEVELOPMENT_VERSION: "1.0.2-SNAPSHOT"
+  RELEASE_VERSION: "1.0.1"
```
tpage-alfresco, line 9: These env variables are only used for releases, so we don't need them here:
```suggestion
```

.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -0,0 +1,87 @@
+name: Veracode and Nightly Tests run
+
+on:
+  schedule:
+    - cron: "0 0 * * *"  # Runs every night at midnight
+
+env:
+  DEVELOPMENT_VERSION: "1.0.2-SNAPSHOT"
+  RELEASE_VERSION: "1.0.1"
+  GITHUB_ACTIONS_DEPLOY_TIMEOUT: 60
```
tpage-alfresco, line 10: We're not going to deploy anything overnight:
```suggestion
```

.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -0,0 +1,87 @@
...
+env:
+  DEVELOPMENT_VERSION: "1.0.2-SNAPSHOT"
+  RELEASE_VERSION: "1.0.1"
+  GITHUB_ACTIONS_DEPLOY_TIMEOUT: 60
+  MAVEN_CLI_OPTS: "-B -e -fae -V -DinstallAtEnd=true -DfailIfNoTests=false -U -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn -Pdistribution "
+  MAVEN_PASSWORD: ${{ secrets.NEXUS_PASSWORD }}
+  MAVEN_USERNAME: ${{ secrets.NEXUS_USERNAME }}
+  JAVA_VERSION: "17"
+  JAVA_VERSION_SUPPORTED_BY_REPO: "11"
+  S3_BUCKET_REGION: "eu-west-1"
```
tpage-alfresco, line 16: I assume this is just used when choosing where to publish the released artifacts to.
```suggestion
```

.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -0,0 +1,96 @@
...
+
+jobs:
+  veracode_sca:
+    name: "Veracode - Source Clear Scan (SCA)"
+    runs-on: ubuntu-latest
+    if: >
+      (github.ref_name == 'master' || startsWith(github.ref_name, 'release/') || github.event_name == 'schedule' || github.actor == 'dependabot[bot]') &&
```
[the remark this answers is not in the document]
PWS-TE: Changed

.github/workflows/nightly_tests_and_veracode.yml
PWS-TE: Deleted

.github/workflows/nightly_tests_and_veracode.yml
PWS-TE: Deleted

.github/workflows/nightly_tests_and_veracode.yml
PWS-TE: Deleted

.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -0,0 +1,83 @@
+name: Veracode and Nightly Tests run
+
+on:
```
tpage-alfresco, line 3: I don't think you should have removed the PR entry here. We want to run SCA against the Dependabot PRs, and they won't be picked up by the schedule. Additionally I just read the docs about "schedule" and it only runs against the default branch (master) unless you pick a different branch. I don't think we can run against all `release/*` branches like this. (This isn't a deal breaker since we have no release branches at all yet!)

PWS-TE: I restored to the previous state with pull_request and push

.github/workflows/nightly_tests_and_veracode.yml
tpage-alfresco: The nightly tests will only run once a day due to `if: github.event_name == 'schedule'` and this will enable the SCA scan to run against Dependabot PRs, which I think is useful. The SAST scan should run against our PRs to check for coding mistakes. I think this looks good now.

tpage-alfresco (approved):
LGTM - thanks @PWS-TE

azakrzewski-hy:
Before we merge this PR we need to make sure that all expected jobs are running during push/pull_request events from. Right now many are skipped, like jobs with e2e tests

PWS-TE:
I checked and in old pipelines it worked the same way as now, i.e. pre_comit, pcm_scan and veracode steps were launched:
https://github.com/Alfresco/hxinsight-connector/actions/runs/12809887035
https://github.com/Alfresco/hxinsight-connector/actions/runs/12808698951

Only during merge all steps were launched:
https://github.com/Alfresco/hxinsight-connector/actions/runs/12808873476

azakrzewski-hy (approved):
Thanks Paweł

PR #827, opened 2025-02-03
CIN-303 Add content digest identifier algorithm

kamilbochno:
Screenshot 2025-02-07 145357

### DIFF
--- a/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/DigestIdentifierParams.java
+++ b/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/DigestIdentifierParams.java
@@ -1,0 +1,50 @@
+/*
+ * #%L
+ * Alfresco HX Insight Connector
+ * %%
+ * Copyright (C) 2023 - 2025 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+
+package org.alfresco.hxi_connector.live_ingester.adapters.messaging.repository.util;
+
+public record DigestIdentifierParams(String digestAlgorithm, String nodeId, String propertyName, String versionNumber)
+{
+    public String getDigestAlgorithm()
+    {
+        return digestAlgorithm;
+    }
+
+    public String getNodeId()
+    {
+        return nodeId;
+    }
+
+    public String getPropertyName()
+    {
+        return propertyName;
+    }
+
+    public String getVersionNumber()
+    {
+        return versionNumber;
+    }
+}
--- a/live-ingester/src/main/resources/application.yml
+++ b/live-ingester/src/main/resources/application.yml
@@ -62,6 +62,7 @@ hyland-experience:
     ingestion:
       base-url: http://localhost:8001
   storage:
+    digest-algorithm: SHA-256
     location:
       endpoint: ${hyland-experience.insight.ingestion.base-url}/presigned-urls?httpMethod=POST&throwExceptionOnFailure=false
   ingester:
--- a/live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/ContentUtilsTest.java
+++ b/live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/ContentUtilsTest.java
@@ -1,0 +1,76 @@
+/*
+ * #%L
+ * Alfresco HX Insight Connector
+ * %%
+ * Copyright (C) 2023 - 2025 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+
+package org.alfresco.hxi_connector.live_ingester.adapters.messaging.repository.util;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ContextConfiguration;
+
+@SpringBootTest
+@ContextConfiguration(classes = {ContentUtils.class})
+class ContentUtilsTest
+{
+    @Value("${hyland-experience.storage.digest-algorithm}")
+    private String digestAlgorithm;
+
+    @Test
+    void shouldThrowExceptionForInvalidAlgorithm()
+    {
+        // given
+        String invalidAlgorithm = "INVALID-11";
+        String nodeId = "02acf462-533d-4e1b-9825-05fa934140da";
+        String propertyName = "cm:content";
+        String versionNumber = "1.0";
+        DigestIdentifierParams params = new DigestIdentifierParams(invalidAlgorithm, nodeId, propertyName, versionNumber);
+
+        // when & then
+        assertThrows(IllegalArgumentException.class, () -> {
+            ContentUtils.generateDigestIdentifier(params);
+        });
+    }
+
+    @Test
+    void shouldGenerateDigestIdentifier()
+    {
+        // given
+        String nodeId = "02acf462-533d-4e1b-9825-05fa934140da";
+        String propertyName = "cm:content";
+        String versionNumber = "1.0";
+        DigestIdentifierParams params = new DigestIdentifierParams(digestAlgorithm, nodeId, propertyName, versionNumber);
+
+        // when
+        String generatedDigestIdentifier = ContentUtils.generateDigestIdentifier(params);
+        String expectedDigestIdentifier = "AKQ0EIuMVIcdLj9vXd/7YJ7WmX6/lMpFIwuW2VsNo6s=";
+
+        // then
+        assertEquals(expectedDigestIdentifier, generatedDigestIdentifier);
+    }
+}
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -285,6 +285,16 @@
         "is_secret": false
       }
     ],
+    "live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/ContentUtilsTest.java": [
+      {
+        "type": "Base64 High Entropy String",
+        "filename": "live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/ContentUtilsTest.java",
+        "hashed_secret": "ac91bf153e9a3b8c7e7b60e011fdd1f630252f17",
+        "is_verified": false,
+        "line_number": 71,
+        "is_secret": false
+      }
+    ],
     "prediction-applier/src/integration-test/java/org/alfresco/hxi_connector/prediction_applier/util/ContainerSupport.java": [
       {
         "type": "Hex High Entropy String",
@@ -296,5 +306,5 @@
       }
     ]
   },
-  "generated_at": "2025-01-17T09:16:04Z"
+  "generated_at": "2025-02-07T12:55:07Z"
 }
--- a/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/ContentUtils.java
+++ b/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/ContentUtils.java
@@ -1,0 +1,56 @@
+/*
+ * #%L
+ * Alfresco HX Insight Connector
+ * %%
+ * Copyright (C) 2023 - 2025 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+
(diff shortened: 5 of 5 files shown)


live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/mapper/property/PropertyMappingHelper.java
```diff
@@ -183,4 +187,26 @@ private static String getUserId(NodeResource node, Function<NodeResource, UserIn
...
+            throw new RuntimeException("Error generating digest", e);
+        }
+    }
+
+    private static Optional<String> getContentDigest(NodeResource nodeResource)
+    {
+        return ofNullable(nodeResource)
+                .map(NodeResource::getContent)
+                .map(ContentInfo::getMimeType)
+                .map(PropertyMappingHelper::generateDigest);
```
tpage-alfresco, line 210: If I've read this correctly then I think we're generating the digest of the mimetype string. This is a good start, but we need to use something that uniquely identifies the piece of content. My suggestion in the ticket is to include the node id, the property name (`cm:content`) and the version number.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/EventUtils.java
```diff
@@ -211,4 +215,23 @@ public static long getEventTimestamp(RepoEvent<DataAttributes<NodeResource>> eve
...
         return event.getTime().toInstant().toEpochMilli();
     }
+
+    public static String generateSHA256Digest(String nodeId, String propertyName, String versionNumber) {
+        try {
+            String input = nodeId + "-" + propertyName + "-" + versionNumber;
+            MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            byte[] hash = digest.digest(input.getBytes(StandardCharsets.UTF_8));
+            return Base64.getEncoder().encodeToString(hash);
+        } catch (NoSuchAlgorithmException e) {
```
tpage-alfresco, line 225: I guess this looks a bit weird at the moment because we're hard coding the algorithm name above. I doubt SHA-256 is going anywhere soon, but you could make the algorithm configurable, and then the exception could inform the admin there was a misconfiguration. I'd suggest doing this as a separate PR though, and I think it will be a useful exercise to help understanding the way properties work in Spring.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/EventUtils.java
```diff
@@ -211,4 +215,23 @@ public static long getEventTimestamp(RepoEvent<DataAttributes<NodeResource>> eve
...
+            throw new RuntimeException("Error generating SHA-256 digest", e);
+        }
+    }
+
+    public static String getPropertyName(RepoEvent<DataAttributes<NodeResource>> event) {
+        return (String) event.getData().getResource().getProperties().get("cm:name");
+    }
+
+    public static String getVersionNumber(RepoEvent<DataAttributes<NodeResource>> event) {
```
tpage-alfresco, line 234: I don't think this code is used.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/hx_insight/model/ContentMetadata.java
```diff
@@ -41,4 +41,5 @@ public class ContentMetadata
     private final String sourceMimeType;
     private final Long size;
     private final String name;
+    private final String digest;
```
tpage-alfresco, line 44: I think the Ingest API calls this the "digest", but we're not creating the digest that would be expected by someone reading this class (the digest of the actual content). I would suggest renaming this to `digestIdentifier` to make it a bit clearer what is contained in the field.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializerTest.java
```diff
@@ -145,11 +150,15 @@ public void canCopeWithNullUsers()
...
+    public void shouldSetContentProperty() throws Exception
     {
+        String expectedDigest = EventUtils.generateSHA256Digest(NODE_ID, "cm:name", "1.0");
+
         UpdateNodeEvent event = new UpdateNodeEvent(NODE_ID, CREATE, SOURCE_ID, TIMESTAMP)
                 .addContentInstruction(new ContentProperty(CONTENT_PROPERTY, "content-id", "application/pdf",
-                        "application/msword", 123L, "something.doc"));
+                        "application/msword", 123L, "something.doc", expectedDigest));
+
+        event.setDigest(expectedDigest);
```
tpage-alfresco, line 161: I don't think this line is needed since it seems to be in the constructor above.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializerTest.java
```diff
@@ -145,11 +150,15 @@ public void canCopeWithNullUsers()
     }
 
     @Test
-    public void shouldSetContentProperty()
+    public void shouldSetContentProperty() throws Exception
     {
+        String expectedDigest = EventUtils.generateSHA256Digest(NODE_ID, "cm:name", "1.0");
```
tpage-alfresco, line 155: I don't think we need a real digest here. Can't we just use something like `"0123456789abcdef"` or `"fake-digest-identifier"`?

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializerTest.java
```diff
@@ -166,24 +175,70 @@ public void shouldSetContentProperty()
                           "content-metadata": {
                             "size": 123,
                             "name": "something.doc",
+                            "digest": "%s",
```
tpage-alfresco, line 178: Similarly, I think we can write the literal value in this json too.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/EventUtilsTest.java
```diff
@@ -55,6 +55,56 @@
...
+        String nodeId = "02acf462-533d-4e1b-9825-05fa934140da";
+        String propertyName = "cm:name";
+        String versionNumber = "1.0";
+
+        // when
+        String digest = EventUtils.generateSHA256Digest(nodeId, propertyName, versionNumber);
+        String expectedDigest = digest;
+
+        // then
+        assertEquals(expectedDigest, digest);
```
tpage-alfresco, line 105: cf. line 102. I think you need to set `expectedDigest` to a literal.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/metadata/IngestNodeCommandHandlerTest.java
```diff
@@ -143,7 +143,7 @@ void shouldSetContentProperty()
             IngestNodeCommand command = new IngestNodeCommand(
                     NODE_ID,
                     CREATE,
-                    Set.of(new ContentPropertyUpdated(CONTENT_PROPERTY, "content-id", "application/pdf", "application/msword", 123L, "something.doc")),
+                    Set.of(new ContentPropertyUpdated(CONTENT_PROPERTY, "content-id", "application/pdf", "application/msword", 123L, "something.doc", "")),
```
tpage-alfresco, line 146: Same comment here. Come up with something similar to `"something.doc"` that explains what the fake value represents.

tpage-alfresco (changes requested):
Great start! If you have any questions then feel free to get in contact, but I think you've worked out all the important things - loading the right properties, how to hash a string, etc.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/hx_insight/model/ContentMetadata.java
kamilbochno: changed to digestIdentifier

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializerTest.java
kamilbochno: changed to fake-digest-identifier

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/metadata/IngestNodeCommandHandlerTest.java
kamilbochno: changed to lieral value fake-digest-identifier

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/EventUtils.java
kamilbochno: Implemented solution with possibility to configure hashing algorithm

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/EventUtilsTest.java
kamilbochno: In this case, where the algorithm generates the appropriate hash, we cannot display it as literal value because it will not pass, so I preferred to do it this way

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/EventProcessor.java
```diff
@@ -2,7 +2,7 @@
  * #%L
  * Alfresco HX Insight Connector
  * %%
- * Copyright (C) 2023 - 2024 Alfresco Software Limited
+ * Copyright (C) 2023 - 2025 Alfresco Software Limited
```
tpage-alfresco, line 5: This shouldn't need to be changed if you're not changing the file.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/EventUtilsTest.java
```diff
@@ -45,15 +46,45 @@
...
+        String nodeId = "02acf462-533d-4e1b-9825-05fa934140da";
+        String propertyName = "cm:content";
+        String versionNumber = "1.0";
+
+        // when
+        String generatedDigestIdentifier = EventUtils.generateDigestIdentifier(nodeId, propertyName, versionNumber);
+        String expectedDigestIdentifier = generatedDigestIdentifier;
+
+        // then
+        assertEquals(expectedDigestIdentifier, generatedDigestIdentifier);
```
tpage-alfresco, line 86: You can't have tests like this - see line 83 above.

tpage-alfresco:
You should need to add something to `rest/hxinsight/requests` for the new request to check whether the file has already been uploaded or not. Apparently it will return some object like `{"answer": true}`

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/EventProcessor.java
kamilbochno: Yes, restored

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/EventUtilsTest.java
kamilbochno: Added false positive to `.secrets.baseline`

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/OpenApiRequestValidationTest.java
```diff
@@ -114,6 +114,20 @@ void testUpdateRequestToIngestionEvents()
         assertThat(schemaValidator.validate(propertiesNode.toString(), propertiesSchema, null).getMessages()).isEmpty();
     }
 
+    @Disabled("This test is disabled because the OpenApi Specification update is not finished yet.")
+    @SneakyThrows
+    @Test
+    void testFileExistsRequestToIngestionEvents()
```
tpage-alfresco, line 120: Let's save this for a future PR and remove it from this one.

live-ingester/src/integration-test/resources/rest/hxinsight/requests/file-exists.yml
```diff
@@ -0,0 +1,72 @@
+url: /v1/ingestion-events
+headers:
```
tpage-alfresco, line 2: Same with this file.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/hx_insight/model/ContentMetadata.java
```diff
@@ -41,4 +41,5 @@ public class ContentMetadata
     private final String sourceMimeType;
     private final Long size;
     private final String name;
+    private final String digestIdentifier;
```
tpage-alfresco, line 44: Let's remove this change for the time being too. Also the test changes caused by this.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/EventUtils.java
```diff
@@ -211,4 +228,19 @@ public static long getEventTimestamp(RepoEvent<DataAttributes<NodeResource>> eve
     {
         return event.getTime().toInstant().toEpochMilli();
     }
+
+    public static String generateDigestIdentifier(String nodeId, String propertyName, String versionNumber)
```
tpage-alfresco, line 232: Let's move this new method to a new utility class called `ContentUtils`

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/EventUtils.java
```diff
@@ -66,6 +71,18 @@ public final class EventUtils
     public static final String PREDICTION_REVIEW_STATUS_PROPERTY = "hxi:reviewStatus";
     public static final String PREDICTION_UNREVIEWED = "UNREVIEWED";
     public static final String PREDICTION_CONFIRMED = "CONFIRMED";
+    private static String digestAlgorithm;
+
+    @Value("${digestIdentifier.algorithm}")
+    public void setDigestIdentifierAlgorithm(String digestAlgorithm)
```
tpage-alfresco, line 77: This won't work because there won't be an instance of this class. Instead let's make the `digestAlgorithm` a parameter of the new method. We'll solve wiring up the Spring configuration in a later PR.

live-ingester/src/main/resources/application.yml
```diff
@@ -15,6 +15,9 @@ application:
   source-id: alfresco-dummy-source-id-0a63de491876
   version: "@project.version@"
 
+digestIdentifier:
+  algorithm: SHA-256
```
tpage-alfresco, line 19: This can stay for the time being - we'll use it in the next PR.

tpage-alfresco: Also let's change the property name to:
```
hyland-experience:
  storage:
    digest-algorithm: SHA-256
```

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/EventUtilsTest.java
```diff
@@ -45,15 +46,45 @@
...
 
+@SpringBootTest
+@ContextConfiguration(classes = {EventUtils.class})
 class EventUtilsTest
 {
+    @Value("${digestIdentifier.algorithm}")
+    private String digestAlgorithm;
+
+    @Test
+    void shouldSetDigestIdentifierAlgorithm()
```
tpage-alfresco, line 67: Instead let's test an invalid algorithm and see if we get an exception.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/DigestIdentifierParams.java
```diff
@@ -0,0 +1,63 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+
+package org.alfresco.hxi_connector.live_ingester.adapters.messaging.repository.util;
+
+public class DigestIdentifierParams
```
tpage-alfresco, line 29: Have a go replacing this class with a `record`. You can see some information about the similarities/differences here: https://www.baeldung.com/java-record-keyword

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/util/ContentUtilsTest.java
```diff
@@ -0,0 +1,76 @@
...
+class ContentUtilsTest
+{
+    @Value("${hyland-experience.storage.digest-algorithm}")
+    private String digestAlgorithm;
+
+    @Test
+    void shouldThrowExceptionForInvalidAlgorithm()
+    {
+        // given
+        String invalidAlgorithm = "SHA-11";
```
tpage-alfresco, line 48: I would make it even more obvious - e.g.
```suggestion
        String invalidAlgorithm = "INVALID-11";
```

PR #831, opened 2025-02-05
[CIN-1802] Disable failing tests from OpenApiRequestValidationTest due to schema update

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-1802

PR to disable failing tests from OpenApiRequestValidationTest because of schema changes in /ingestion-events endpoint are not finished yet.
Tests should be enabled again after update of the OpenAPI Specification will be finished completely


kamilbochno (approved):

tpage-alfresco (approved):
Thanks Adam :+1:

PR #853, opened 2025-02-20
CIN-2018: specify data types with non annotated properties

kamilbochno:


kamilbochno:
Currently, the `OpenApiRequestValidationTest` file doesn't pass due to the difference with the Ingestion API schema. I contacted team responsible for Ingestion API and they are currently implementing type schema.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
```diff
@@ -135,6 +117,55 @@ private void writeProperty(JsonGenerator jgen, FieldType fieldType, String name,
...
+            break;
+        case VERSION_TYPE:
+        case ALLOW_ACCESS:
+        case DENY_ACCESS:
+        case TITLE:
+        case VERSION_LABEL:
+            jgen.writeObjectField("type", "string");
+            break;
+        default:
+            break;
```
tpage-alfresco, line 165: This won't really work in a real system. The customer will define a long list of extra properties and we won't know about them ahead of time. I suggest that as a first version we use the property value to attempt to guess the type and assume `string` if we can't tell.

In future we could load the content model from ACS and then use this to determine the property types. I think this will be considerably more work though, so I'm suggesting the quick version as a starting point.

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/MatchingContentMappingRequestIntegrationTest.java
```diff
@@ -124,8 +124,8 @@ void givenMappingForImage_whenContentWithMatchingTypeIngested_thenProcessWithTra
                           }
                         }
                       },
-                      "ALLOW_ACCESS": {"value": ["GROUP_EVERYONE"]},
-                      "DENY_ACCESS": {"value": []}
+                      "ALLOW_ACCESS": {"type": "string", "value": ["GROUP_EVERYONE"]},
+                      "DENY_ACCESS": {"type": "string", "value": []}
```
tpage-alfresco, line 128: This raises an interesting question - how can we guess the type for an empty array. Anyway, for now I suggest we assume `string` unless we can immediately tell otherwise from the value.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
kamilbochno: Changed to a more flexible implementation

live-ingester/src/integration-test/resources/rest/hxinsight/requests/create-or-update-document.yml
```diff
@@ -35,16 +36,19 @@ body: [
...
       "cm:versionLabel": {
+        "type": "string",
         "value": "1.0"
       },
       "createdBy": {
         "value": "admin",
         "annotation": "createdBy"
       },
       "ALLOW_ACCESS": {
+        "type": "object",
```
tpage-alfresco, line 51: Arrays of strings should be `string`. The empty array for `DENY_ACCESS` should also be `string`, but it's not easy to determine this. My suggestion is to strip all empty arrays from the ingest message for the time being.

live-ingester/src/integration-test/resources/rest/hxinsight/requests/upload-references-document.yml
```diff
@@ -12,6 +12,7 @@ body: [
     "sourceTimestamp": 1611656982995,
     "properties": {
       "cm:content": {
+        "type": "object",
```
tpage-alfresco, line 15: I don't think there should be a `type` for `file` references.

PR #863, opened 2025-02-26
[CIN-2018]: specify data types with non annotated properties

kamilbochno:
Screenshot 2025-03-06 152745

After updating schema from Ingestion API side, all tests including OpenApi schema test are now passing


live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/MatchingContentMappingRequestIntegrationTest.java
```diff
@@ -124,8 +124,8 @@ void givenMappingForImage_whenContentWithMatchingTypeIngested_thenProcessWithTra
                           }
                         }
                       },
-                      "ALLOW_ACCESS": {"value": ["GROUP_EVERYONE"]},
-                      "DENY_ACCESS": {"value": []}
+                      "ALLOW_ACCESS": {"type": "string", "value": ["GROUP_EVERYONE"]},
+                      "DENY_ACCESS": {"type": "string", "value": []}
```
tpage-alfresco, line 128: We need to remove these empty arrays (as discussed).

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
```diff
@@ -135,6 +114,65 @@ private void writeProperty(JsonGenerator jgen, FieldType fieldType, String name,
...
+        default:
+            hasAnnotation = false;
+            break;
+        }
+        return hasAnnotation;
+    }
+
+    private void writeType(JsonGenerator jgen, String name, Object value) throws IOException
+    {
+        if ("cm:content".equals(name))
```
tpage-alfresco, line 152: We should work via the value instead. `cm:content` will always be a file reference if it's present.

kamilbochno: Changed to value-based check

PWS-TE (approved):

PWS-TE:
Looks 🆗

PWS-TE (approved):

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
```diff
@@ -135,6 +114,65 @@ private void writeProperty(JsonGenerator jgen, FieldType fieldType, String name,
...
+        }
+        else if (value instanceof Integer)
+        {
+            type = "integer";
+        }
+        else if (value instanceof Float)
+        {
+            type = "float";
+        }
+        else
```
tpage-alfresco, line 169: I think you should also cope with collections. You could refactor this to extract a function:
```
    private String selectTypeByValue(Object value)
    {
        if (value instanceof Boolean)
            ...
        else if (value instanceof Collection)
        {
            Collection collection = (Collection) value;
            if (collection.isEmpty())
            {
                // This should not happen because empty collections should already be filtered out.
                throw new ...;
            }
            return selectTypeByValue(collection.stream().findAny());
        }
        // Default to string.
        return "string";
    }
```

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/MatchingContentMappingRequestIntegrationTest.java
krdabrowski: I'm not convinced to empty string inside an array instead of empty array... what is the reason behind it?

kamilbochno: I think that after adding types, in this case a string, there are no empty arrays without type specification. According to the updated Ingestion API schema, an empty array is currently an invalid data type. https://github.com/HylandSoftware/hxai-insight-integrations-ingestion-api/blob/main/src/tests/data/ingestion-events/invalid/data-types/value-string-array-empty.json

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/UpdateRequestIntegrationTest.java
```diff
@@ -794,7 +810,7 @@ void testUpdatePermissions()
                       "GROUP_EVERYONE",
                       "abeecher"
                     ],
-                    "resourceDeniedAuthorities": []
+                    "resourceDeniedAuthorities": [ "" ]
```
tpage-alfresco, line 813: We shouldn't be changing the input value to be populated. When the `DENY_ACCESS` list is empty then we decided that the property should be omitted from the output.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/mapper/property/PropertyMappingHelper.java
```diff
@@ -122,7 +123,7 @@ public static Optional<PropertyDelta<?>> calculateDenyAccessDelta(RepoEvent<Data
 
         if (eventData.getResourceDeniedAuthorities() == null)
         {
-            return Optional.of(PropertyDelta.updated(DENY_ACCESS, Set.of()));
+            return Optional.of(PropertyDelta.updated(DENY_ACCESS, Set.of(EMPTY_STRING)));
```
tpage-alfresco, line 126: We should update the two methods here that are dealing with lists so that we handle the case where there's an empty list better - `calculateAllowAccessDelta` and `calculateDenyAccessDelta`.

live-ingester/src/integration-test/java/org/alfresco/hxi_connector/live_ingester/domain/usecase/e2e/repository/UpdateRequestIntegrationTest.java
kamilbochno: Changed

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/mapper/property/PropertyMappingHelper.java
kamilbochno: Applied

kamilbochno:
> This is great - thanks. There are already plenty of tests for this, but the one extra thing I suggest you should consider is writing some unit tests for the different flows in `selectTypeByValue`. In order to do this you need to change the scope from `private` (e.g. to package protected) so that the unit test can access it.

I agree, I will add additional unit tests for `selectTypeByValue`. I think I can do it in the next PR as discussed.

PR #879, opened 2025-03-05
[CIN-2358] Fix build for nightly tests job

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-2358

### DIFF
--- a/.github/workflows/nightly_tests_and_veracode.yml
+++ b/.github/workflows/nightly_tests_and_veracode.yml
@@ -80,16 +80,10 @@ jobs:
     steps:
       - name: "Checkout repository"
         uses: actions/checkout@v4
-      - name: "Free up hosted runner disk space"
-        uses: Alfresco/alfresco-build-tools/.github/actions/free-hosted-runner-disk-space@v8.14.1
-      - name: "Login to Quay.io"
-        uses: docker/login-action@v3
-        with:
-          registry: quay.io
-          username: ${{ secrets.QUAY_USERNAME }}
-          password: ${{ secrets.QUAY_PASSWORD }}
       - name: "Set up JDK"
         uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v8.14.1
+      - name: "Build SNAPSHOT dependencies"
+        run: mvn ${{ env.MAVEN_CLI_OPTS }} clean install -U -DskipTests -pl alfresco-hxinsight-connector-common-authentication,alfresco-hxinsight-connector-common-test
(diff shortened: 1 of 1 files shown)


azakrzewski-hy:
In this PR, feature/** in branches and pull_request condition instead of schedule are used for demonstration purpose only

PWS-TE (approved):
Looks good 👍

kamilbochno (approved):

.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -74,14 +71,16 @@ jobs:
...
     runs-on: ubuntu-latest
-    if: github.event_name == 'schedule'
+    if: github.event_name == 'pull_request'
     steps:
       - name: "Checkout repository"
         uses: actions/checkout@v4
       - name: "Set up JDK"
         uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v8.14.1
+      - name: "Build SNAPSHOT dependencies"
+        run: mvn ${{ env.MAVEN_CLI_OPTS }} clean install -DskipTests -pl common-authentication,common-test
```
tpage-alfresco, line 81: You should use `-am` here too so that any other necessary snapshots are built. In fact you can just use: `-pl common-authentication -am`

.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -74,14 +71,16 @@ jobs:
...
       - name: "Build application"
-        run: mvn clean install -DskipTests -pl live-ingester
+        run: mvn ${{ env.MAVEN_CLI_OPTS }} clean install -DskipTests
```
tpage-alfresco, line 83: Oh. Ok, well this builds everything, making it pointless to do the previous step. I think we should get rid of "Build SNAPSHOT dependencies" and just do:
```suggestion
        run: mvn ${{ env.MAVEN_CLI_OPTS }} clean install -DskipTests -pl live-ingester -am
```

.github/workflows/nightly_tests_and_veracode.yml
azakrzewski-hy: this will be deleted as mentioned in your next comment

.github/workflows/nightly_tests_and_veracode.yml
azakrzewski-hy: thanks, change is implemented

tpage-alfresco (approved):
LGTM!

PR #886, opened 2025-03-10
[CIN-2351] Set Veracode SAST scan for project code only

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-2351

feature/** in branches and pull_request condition instead of schedule are used for demonstration purpose only - its temporary and will be removed after acceptance

We should not perform any Veracode SAST scan on code which we don't maintain (e.g. coming from an external library), it should scan only jxinsight-connector modules

### DIFF
--- a/.github/workflows/nightly_tests_and_veracode.yml
+++ b/.github/workflows/nightly_tests_and_veracode.yml
@@ -3,6 +3,7 @@ name: Veracode and Nightly Tests run
 on:
   pull_request:
     branches:
+      - feature/**
       - master
       - release/**
 
@@ -36,7 +37,7 @@ jobs:
     name: "Pipeline SAST Scan"
     runs-on: ubuntu-latest
     if: >
-      (github.ref_name == 'master' || startsWith(github.ref_name, 'release/') || github.event_name == 'schedule') &&
+      (github.ref_name == 'master' || startsWith(github.ref_name, 'release/') || github.event_name == 'pull_request') &&
       github.actor != 'dependabot[bot]' &&
       !contains(github.event.head_commit.message, '[skip build]')
     steps:


.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -64,6 +67,11 @@ jobs:
           summary_output_file: results.json
           summary_display: true
           baseline_file: baseline.json
+          include: "alfresco-hxinsight-connector-bulk-ingester-*,
+          alfresco-hxinsight-connector-hxinsight-extension-*,
+          alfresco-hxinsight-connector-live-ingester-*,
+          alfresco-hxinsight-connector-prediction-applier-*,
+          alfresco-hxinsight-connector-distribution-*"
```
tpage-alfresco, line 74: Can we instead use `alfresco-hxinsight-connector*` to cover all these modules and any others that may be created by this project in the future?

.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -15,6 +16,8 @@ on:
 
 env:
   MAVEN_CLI_OPTS: "-B -e -fae -V -DinstallAtEnd=true -DfailIfNoTests=false -U -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn -Pdistribution "
+  MAVEN_PASSWORD: ${{ secrets.NEXUS_PASSWORD }}
+  MAVEN_USERNAME: ${{ secrets.NEXUS_USERNAME }}
```
gionn, line 20: you should be able to move this to `Build application` step only to avoid giving too wide access (especially because we are running the upstream veracode action here)

.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -36,7 +39,7 @@ jobs:
     name: "Pipeline SAST Scan"
     runs-on: ubuntu-latest
     if: >
-      (github.ref_name == 'master' || startsWith(github.ref_name, 'release/') || github.event_name == 'schedule') &&
+      (github.ref_name == 'master' || startsWith(github.ref_name, 'release/') || github.event_name == 'pull_request') &&
```
gionn, line 42: this was just for testing right? should be reverted

you could add a workflow_dispatch trigger for the future (and add an additional `|| github.event_name == 'workflow_dispatch'` here)

.github/workflows/nightly_tests_and_veracode.yml
azakrzewski-hy: change implemented in https://github.com/Alfresco/hxinsight-connector/pull/886/commits/b3cfa9731e0fc9b7f52a350b5e9995aadf4f499c

.github/workflows/nightly_tests_and_veracode.yml
azakrzewski-hy: secrets moved into build application step in: https://github.com/Alfresco/hxinsight-connector/pull/886/commits/b3cfa9731e0fc9b7f52a350b5e9995aadf4f499c

gionn (approved):

tpage-alfresco (approved):
Other than the branch and trigger condition (which you already flagged) then this looks good. Thanks Adam! :+1:

PWS-TE (approved):

PR #888, opened 2025-03-11
[CIN-2018] Add new test cases to check data types with non annotated properties

kamilbochno:
Screenshot 2025-03-11 094704

### DIFF
--- a/live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializerTest.java
+++ b/live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializerTest.java
@@ -26,7 +26,12 @@
 
 package org.alfresco.hxi_connector.live_ingester.adapters.config.jackson;
 
+import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.anyString;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
 
 import static org.alfresco.hxi_connector.common.constant.NodeProperties.CONTENT_PROPERTY;
 import static org.alfresco.hxi_connector.common.constant.NodeProperties.CREATED_AT_PROPERTY;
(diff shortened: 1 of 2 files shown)


live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializerTest.java
```diff
@@ -185,6 +190,43 @@ public void shouldOnlySendUpdatedContentMetadata()
...
+        Exception exception = assertThrows(IllegalArgumentException.class, () -> {
+            serializer.selectTypeByValue(Collections.emptyList());
+        });
+        assertEquals("Empty collections should not be passed to selectTypeByValue.", exception.getMessage());
+    }
+
+    @Test
+    public void shouldReturnStringForMixedTypeCollection()
+    {
+        assertEquals("string", serializer.selectTypeByValue(Arrays.asList(1, "test", 3.14)));
```
tpage-alfresco, line 215: I don't see how this works. The code looks like it should select an arbitrary entry from the list and use that:
https://github.com/Alfresco/hxinsight-connector/pull/888/files#diff-d1784b0b5262c967b22e4bfffe85820bb1d1bc0241e354dcd92f48c4776b2250R187

I hadn't thought about mixed type collections (I'm not sure whether it's possible in Alfresco with `d:any`), but regardless this is a neat idea for a test.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializerTest.java
```diff
@@ -185,6 +190,43 @@ public void shouldOnlySendUpdatedContentMetadata()
...
+    }
+
+    @Test
+    public void shouldReturnStringForNullValue()
+    {
+        assertEquals("string", serializer.selectTypeByValue(null));
+    }
+
+    @Test
+    public void shouldReturnStringForCollectionWithDifferentTypes()
```
tpage-alfresco, line 225: This looks like a duplicate test of `shouldReturnStringForMixedTypeCollection`.

tpage-alfresco (changes requested):
I think we should also have some mock-based tests for `writeProperty` - in particular testing an empty and a populated array.

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializerTest.java
```diff
@@ -29,22 +29,33 @@
 import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.*;
```
tpage-alfresco, line 32: Please avoid wildcard imports. You can do this in IntelliJ by setting the limit to some large number: 
image

tpage-alfresco: Wildcard imports could potentially lead to unexpected behaviour if a class is added/removed from the target library and its name clashes with another class.

tpage-alfresco (approved):
Looks good to me! Thanks Kamil :+1:

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializerTest.java
kamilbochno: Deleted

live-ingester/src/test/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializerTest.java
kamilbochno: Wildcard is now deleted. Replaced to single specified imports.

PR #889, opened 2025-03-11
[CIN-2351] Set SAST to fail and run on every PR

azakrzewski-hy:
Veracode SAST scan should fail the action check and run on every PR

Changes inside:
- SAST scan is moved to ci.yml file to run on every PR
- When scan finds any issue then GitHub check will fail and summary will be displayed in checked action logs
- When scan does not find any issue it will not display any summary output in the PR console by github-actions bot
- Action github-download-file was removed together with the parameter: baseline_file - as this should be used to filter flows so no need to have it by default till the need arise
- Parameters: summary_output and summary_output_file were removed as not see much value in storing it - when SAST scan fails it will have issue details displayed in checked action logs
- To prevent the PR comments (by github-actions bot), we use the token parameter with an empty value. When the token is not provided, the action is not able to create PR comments.


.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -68,6 +67,12 @@ jobs:
           summary_display: true
           baseline_file: baseline.json
           include: "alfresco-hxinsight-connector*"
+      - name: "Notify on Veracode SAST scan failure"
+        if: failure()
+        uses: Alfresco/alfresco-build-tools/.github/actions/send-teams-notification@v8.14.1
+        with:
+          webhook-url: ${{ secrets.MS_TEAMS_WEBHOOK_URL }}
+          message: "Nightly Veracode SAST scan failed"
```
tpage-alfresco, line 75: I think that this also runs on push and pull requests to master.

azakrzewski-hy: here we expect that notification will be sent for "if" conditions. Same notification action is used for nightly_tests

azakrzewski-hy: notification condition is updated to run for schedule event in:
https://github.com/Alfresco/hxinsight-connector/pull/889/commits/755aa8416e89b5c4bc68591c06c7421492f79e08

gionn: teams notification make sense only for SCA while it runs in scheduled workflow

for SAST as it's related to application code should report any failure during PRs (may need to handle a base list for known issues initially)

azakrzewski-hy: Veracode SAST job was moved to ci.yml in order to run on every PR. 
Work on SCA job will be done in separate PR as it is getting an error:
https://github.com/Alfresco/hxinsight-connector/actions/runs/13800635533

PWS-TE (approved):
Looks ok.

.github/workflows/nightly_veracode_sca_and_openapi_tests_ci.yml
gionn, line 1: avoid renaming workflow files when possible because they interrupt the history and create a new entry in the actions page.

.github/workflows/nightly_veracode_sca_and_openapi_tests_ci.yml
```diff
@@ -1,92 +1,53 @@
...
-        run: mvn ${{ env.MAVEN_CLI_OPTS }} clean install -DskipTests -pl live-ingester -am
-      - name: "Run OpenApi Specification tests"
-        run: mvn test -Dtest=OpenApiRequestValidationTest -pl live-ingester
-      - name: "Notify on failure"
-        if: failure()
-        uses: Alfresco/alfresco-build-tools/.github/actions/send-teams-notification@v8.14.1
-        with:
-          webhook-url: ${{ secrets.MS_TEAMS_WEBHOOK_URL }}
-          message: "Nightly tests failed"
+name: Nightly Veracode SCA and OpenApi tests CI
```
gionn, line 1: see how it render on the actions tab, it's usually better to keep it short

.github/workflows/nightly_veracode_sca_and_openapi_tests_ci.yml
```diff
@@ -1,92 +1,53 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Alfresco/alfresco-build-tools/.github/actions/free-hosted-runner-disk-space@v8.14.1
+      - uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v8.14.1
+      - uses: Alfresco/alfresco-build-tools/.github/actions/veracode@v8.14.1
+        continue-on-error: true
+        with:
+          srcclr-api-token: ${{ secrets.SRCCLR_API_TOKEN }}
+
+  nightly_tests:
```
gionn, line 35: not a big deal, but I would have made two separate workflows as the two jobs doesn't really have much in common (you are just reusing the schedule trigger?)

azakrzewski-hy: yes, you are right, one job is for Veracode scan and another is for tests

.github/workflows/nightly_veracode_sca_and_openapi_tests_ci.yml
azakrzewski-hy: following comment:
https://github.com/Alfresco/hxinsight-connector/pull/889#discussion_r1993017116
we need to change the name anyway for two separate workflows

.github/workflows/nightly_veracode_sca_and_openapi_tests_ci.yml
azakrzewski-hy: following the comment:
https://github.com/Alfresco/hxinsight-connector/pull/889#discussion_r1993017116
when we create two separate workflows, names will be shorter

.github/workflows/nightly_veracode_sca_and_openapi_tests_ci.yml
azakrzewski-hy: ci.yml workflow contains jobs that do not have much in common as well, so if we go that way to make separate workflow for every distinct job then we need to split ci.yml file too. What do you think @tpage-alfresco ?

.github/workflows/nightly_veracode_sca_and_openapi_tests_ci.yml
azakrzewski-hy: name of the workflow was reverted to the one we have on master:
Veracode and Nightly Tests run
if we decide to split the workflows insight the file we'll do it in separate PR as right now there is another PR that fixes on error in that workflow :
https://github.com/Alfresco/hxinsight-connector/pull/892

.github/workflows/nightly_veracode_sca_and_openapi_tests_ci.yml
azakrzewski-hy: if we decide to split the jobs of the workflow ( maybe for ci.yml too ) we will do it in separate PR as right now there is another PR that fixes on error in that workflow :
https://github.com/Alfresco/hxinsight-connector/pull/892

.github/workflows/ci.yml
```diff
@@ -45,6 +45,36 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs:
+      - pre_commit
+    if: >
+      (github.ref_name == 'master' || startsWith(github.ref_name, 'release/') || github.event_name == 'pull_request') &&
+      github.actor != 'dependabot[bot]' &&
+      !contains(github.event.head_commit.message, '[skip build]')
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v8.14.1
```
tpage-alfresco, line 59: You removed the use of the baseline file - was this deliberate? Even though the baseline is currently empty then I would expect it to be useful in the future.

azakrzewski-hy: yes, it was removed deliberately as mentioned:
https://github.com/Alfresco/hxinsight-connector/pull/889#issue-2910288865
I can restore it if we would like to keep it

azakrzewski-hy: github-download-file action and baseline_file parameter is restored for the future use in:
https://github.com/Alfresco/hxinsight-connector/pull/889/commits/755bbad9b8f77cf97b7c0d33635e06469bf418e4

tpage-alfresco (approved):
LGTM

gionn (approved):

PR #893, opened 2025-03-14
[CIN-2468] Convert known dates from unix timestamps to ISO format

kamilbochno:
Due to changes in the Ingestion API schema, I have prepared a PR with changes in dates and a change in format from unix timestamps to ISO format. Additionally, I have added the type required in the updated Ingestion API schema `"type": "datetime"` to the `createdAt` and `modifiedAt` properties.


bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoNodeMapper.java
```diff
@@ -111,12 +111,12 @@ public IngestEvent map(AlfrescoNode alfrescoNode)
                 timeProvider.getCurrentTimestamp());
     }
 
-    private long getCreatedAt(AlfrescoNode alfrescoNode)
+    private String getCreatedAt(AlfrescoNode alfrescoNode)
```
tpage-alfresco, line 114: I can't think of a reason that this isn't also being called for `getModifiedAt`. I think these are the two dates that we're confident of receiving and sending on.

tpage-alfresco (approved):
This looks good to me.

I don't think specifying the `type` actually helps at all, but it's not particularly harmful.

During this review I noticed that we're missing the modification date from bulk ingestion events. We should add this, but it would make sense as a separate PR (even a separate ticket).

azakrzewski-hy (approved):
LGTM

bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoNodeMapper.java
kamilbochno: Yes, it will be done in separate PR.

PR #902, opened 2025-03-20
[CIN-2468] Remove datetime type from createdAt and modifiedAt properties

kamilbochno:
Due to the `datetime` type not being implemented in the Ingestion API yet, we decided to remove the `datetime` type for now.


tpage-alfresco (approved):
Thanks Kamil!

PR #908, opened 2025-03-25
[CIN-2695] Veracode SCA fix for dependabot

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-2695

${{ secrets.SRCCLR_API_TOKEN }} is not available in veracode-sca job when its triggered by the dependabot actor, same secret is available when the job runs on schedule

### DIFF
--- a/.github/workflows/nightly_tests_and_veracode.yml
+++ b/.github/workflows/nightly_tests_and_veracode.yml
@@ -1,5 +1,8 @@
 name: Veracode and Nightly Tests run
 
+permissions:
+  contents: read
+
 on:
   pull_request:
     branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly_tests_and_veracode.yml
```diff
@@ -20,6 +23,9 @@ jobs:
   veracode_sca:
     name: "Veracode - Source Clear Scan (SCA)"
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
```
tpage-alfresco, line 27: Do we need this in two places?

azakrzewski-hy: we can get rid of top-level permissions so only one job will have the explicit permissions

azakrzewski-hy: top level permission is deleted so now nightly_tests will run with default permissions

tpage-alfresco (approved):

PR #909, opened 2025-03-27
[CIN-2695] Fix for dependabot Veracode SCA

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-2695

${{ secrets.SRCCLR_API_TOKEN }} is not available in veracode-sca job when its triggered by the dependabot actor, same secret is available when the job runs on schedule

id-token: write -> fetch an OpenID Connect (OIDC) token

### DIFF
--- a/.github/workflows/nightly_tests_and_veracode.yml
+++ b/.github/workflows/nightly_tests_and_veracode.yml
@@ -3,13 +3,8 @@ name: Veracode and Nightly Tests run
 on:
   pull_request:
     branches:
-      - master
-      - release/**
+      - feature/**
 
-  push:
-    branches:
-      - master
-      - release/**
   schedule:
     - cron: "0 0 * * *"  # Runs every night at midnight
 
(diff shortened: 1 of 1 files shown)


tpage-alfresco:
Reading up on this then it seems that secrets being unavailable is deliberate to avoid a security risk: https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544

Since we're using Veracode for vulnerability scanning then I'd guess we'd actually prefer the risk of this particular secret being exposed, but I've no idea if this is possible without also putting other secrets at the same risk.

PWS-TE (approved):

azakrzewski-hy:
> Reading up on this then it seems that secrets being unavailable is deliberate to avoid a security risk: [dependabot/dependabot-core#3253 (comment)](https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544)
> 
> Since we're using Veracode for vulnerability scanning then I'd guess we'd actually prefer the risk of this particular secret being exposed, but I've no idea if this is possible without also putting other secrets at the same risk.

Thank you Tom, this is very useful information. So as they made the dependabot restricted to not be able to use repository secrets then maybe we could do veracode_sca scan only on schedule time as manipulating an access to secret for dependabot introduces some risk. What do you think ?

tpage-alfresco (approved):
LGTM

PR #910, opened 2025-03-27
[CIN-2773] Property type is failing OpenApi validation

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-2773

"cm:contentPropertyName": {
                "value": {
                    "namespaceURI": "http://www.alfresco.org/model/content/1.0",
                    "localName": "content",
                    "prefixString": "content"
                },
                "type": "string"
            },
is failing OpenApi validation with response:

[400 Bad Request response.json](https://github.com/user-attachments/files/19485152/400.Bad.Request.response.json)

Idea is to fail a build in order to implement a fix for the issue


azakrzewski-hy:
this PR will not be used

PR #911, opened 2025-03-27
CIN-2773 Handle object properties.

tpage-alfresco:
For example the standard property cm:contentPropertyName.

### DIFF
--- a/distribution/src/main/resources/docker-compose/docker-compose.yml
+++ b/distribution/src/main/resources/docker-compose/docker-compose.yml
@@ -299,7 +299,7 @@ services:
         condition: service_healthy
     environment:
       JAVA_TOOL_OPTIONS: -agentlib:jdwp=transport=dt_socket,address=*:5007,server=y,suspend=n
-      LOGGING_LEVEL_ORG_ALFRESCO: INFO
+      LOGGING_LEVEL_ORG_ALFRESCO: DEBUG
       SPRING_ACTIVEMQ_BROKERURL: nio://activemq:61616
       ALFRESCO_REPOSITORY_BASEURL: http://alfresco:8080/alfresco
       ALFRESCO_TRANSFORM_SHAREDFILESTORE_BASEURL: http://shared-file-store:8099
--- a/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
+++ b/live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
@@ -110,7 +110,7 @@ public class UpdateNodeEventSerializer extends StdSerializer<UpdateNodeEvent>
 
             if (value instanceof Map<?, ?> nestedMap)
             {
-                writeNestedProperties(jgen, nestedMap);
+                writeNestedMap(jgen, nestedMap);
             }
             else
             {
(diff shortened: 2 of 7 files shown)


distribution/src/main/resources/docker-compose/docker-compose.yml
```diff
@@ -299,7 +299,7 @@ services:
         condition: service_healthy
     environment:
       JAVA_TOOL_OPTIONS: -agentlib:jdwp=transport=dt_socket,address=*:5007,server=y,suspend=n
-      LOGGING_LEVEL_ORG_ALFRESCO: INFO
+      LOGGING_LEVEL_ORG_ALFRESCO: DEBUG
```
tpage-alfresco, line 302: Since this is used primarily for dev work then it seems useful to set it to DEBUG so we can easily view the events received and sent by the live ingester.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
```diff
@@ -174,23 +201,69 @@ else if (value instanceof Integer)
...
+            jgen.writeObjectFieldStart(key);
+
+            String type = determineType(nestedValue);
+            jgen.writeStringField("type", type);
+
+            if ("object".equals(type))
+            {
+                writeNestedProperties(jgen, (Map<?, ?>) nestedValue);
+            }
+            else if ("array".equals(type) && nestedValue instanceof Collection<?> collection)
```
tpage-alfresco, line 237: I don't think we should ever write `"array"` according to the spec. From what I remember then an array of strings should have the type `"string"`, but I can't see this in the spec now.

tpage-alfresco: Found it here: https://hyland.atlassian.net/wiki/spaces/cin/pages/1817051141/ADR+0006+-+Data+format+for+Objects

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
```diff
@@ -174,23 +201,59 @@ else if (value instanceof Integer)
...
     }
 
-    private String getLowerCase(Object object)
+    private void writeNestedProperties(JsonGenerator jgen, Map<?, ?> nestedMap) throws IOException
     {
-        return object.toString().toLowerCase(ENGLISH);
+        writeMap(jgen, nestedMap, "object");
+    }
+
+    private void writeValue(JsonGenerator jgen, String key, Object value) throws IOException
```
tpage-alfresco, line 227: What's the difference between this and `writeProperty`? I was expecting the recursion to be at a higher level than this.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
```diff
@@ -174,23 +201,59 @@ else if (value instanceof Integer)
...
+        else
+        {
+            jgen.writeStringField("value", value.toString());
+            jgen.writeStringField("type", determineType(value));
+        }
+
+        jgen.writeEndObject();
+    }
+
+    private void writeMap(JsonGenerator jgen, Map<?, ?> map, String type) throws IOException
```
tpage-alfresco, line 249: This is only ever called with `type="object"`, which makes sense, since a map is always an object.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
kamilbochno: refactored to `writeNestedMap` method

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/jackson/UpdateNodeEventSerializer.java
kamilbochno: refactored

tpage-alfresco:
I think this is working correctly now - I'm happy to merge it when you are @kamilbochno

kamilbochno (approved):
LGTM

PR #924, opened 2025-04-15
[CIN-2503] Switch from default into advance setup of GHAS

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-2503

Switch from default into advance setup of GHAS in hxinsight-connector


azakrzewski-hy:
this PR will not be used

PR #926, opened 2025-04-16
fix(CIN-3261): change structure of acs references for answers

kamilbochno:


azakrzewski-hy:
It would be good to rename the brach to have "feature/" or "fix/" prefix and this could allow to trigger tests on CI run, now we skip all tests in this PR. We can expect that before we update WireMock answer tests from QuestionsAndAnswersE2eTest.class should fail with changes of the PR

PR #949, opened 2025-05-28
feat(CIN-4230): add new tests to increase code coverage

kamilbochno:


azakrzewski-hy (approved):
Its good, only one note about repeated variables to consider

tpage-alfresco:
Nice work @kamilbochno!

PR #952, opened 2025-06-02
[CIN-4376] Mitigate false positive for httpclient5 in hxinsight-extension

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-4376

hxinsight-extension depends on ACS and uses libraries from there. In dependency tree for this module, there is: httpclient5:jar:5.4.1:compile 

live-ingester, bulk-ingester and prediction-applier uses: httpclient5-5.4.3.jar
Vulnerability was fixed in 5.4.3 

Run on push is triggered only for testing in this PR


tpage-alfresco (approved):
LGTM

PR #971, opened 2025-06-06
CIN-4452 Add junit-platform which was removed by Spring Boot 3.5.0 upgrade.

tpage-alfresco:


azakrzewski-hy (approved):

kamilbochno (approved):

tpage-alfresco:
We may want this dependency in the other modules too - I didn't check. We noticed this issue in the TCK tests because the build fails when zero tests run. Some of the other parts of the build don't have this restriction and so it's possible we're silently ignoring some tests.

tpage-alfresco:
The code coverage report suggests there are other tests that aren't running too.

krdabrowski:
@tpage-alfresco you could try to place this dependency in `common-test` and `extension` modules instead - maybe this will increase test coverage score 🤔

PR #973, opened 2025-06-09
[CIN-4441] Update flaky test

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-4441

### DIFF
--- a/e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java
+++ b/e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/UpdateNodeE2eTest.java
@@ -198,8 +198,8 @@ public class UpdateNodeE2eTest
             assertThat(actualNode2.properties())
                     .containsKey(PROPERTY_TO_UPDATE)
                     .extracting(map -> map.get(PROPERTY_TO_UPDATE)).isEqualTo(USER_VALUE);
+            WireMock.verify(exactly(0), anyRequestedFor(urlEqualTo("/ingestion-events")));
         }, DELAY_MS);
-        WireMock.verify(exactly(0), anyRequestedFor(urlEqualTo("/ingestion-events")));
     }
 
     @Test


azakrzewski-hy:
Checking what request we are getting in //then part of the test ( when test fails on:
WireMock.verify(exactly(0), anyRequestedFor(urlEqualTo("/ingestion-events"))); )
by adding logging ( before mentioned verify line ):
List<LoggedRequest> requests = WireMock.findAll(anyRequestedFor(urlEqualTo("/ingestion-events")));
        System.out.println("Requests after reset: " + requests.size());
        requests.forEach(req -> System.out.println("This is the request: " + req.getBodyAsString()));
        WireMock.verify(exactly(0), anyRequestedFor(urlEqualTo("/ingestion-events")));

We could see the output:
Requests after reset: 1
This is the request: [{"objectId":"6b735de2-f48d-4c05-b35d-e2f48dbc053c","sourceId":"a1f3e7c0-d193-7023-ce1d-0a63de491876","eventType":"createOrUpdate","sourceTimestamp":1749540213308,"properties":{"usr:accountLocked":{"value":false,"type":"boolean"},"usr:credentialsExpire":{"value":false,"type":"boolean"},"cm:name":{"value":"6b735de2-f48d-4c05-b35d-e2f48dbc053c","annotation":"name"},"usr:username":{"value":"admin","type":"string"},"usr:hashIndicator":{"value":["bcrypt10"],"type":"string"},"usr:salt":{"value":"SENSITIVE_DATA_REMOVED","type":"string"},"usr:accountExpires":{"value":false,"type":"boolean"},"usr:passwordHash":{"value":"SENSITIVE_DATA_REMOVED","type":"string"},"usr:enabled":{"value":true,"type":"boolean"},"type":{"value":"usr:user","annotation":"type"}}}]

This is request is not related to our updatedNode and is likely comes from ingestion events that are sent independently of our update process checked in the test. 
Instead of asserting that there are exactly 0 requests to /ingestion-events, we should filter out irrelevant requests and only verify that no requests related to our test node "updateNode.id()" are present. This way, the test only fails if an ingestion event for our updated node is sent, ignoring unrelated background events.

PWS-TE (approved):

PR #980, opened 2025-06-12
feat(CIN-4547): replace jacoco-report with our python script

kamilbochno:
After implementing the Python script, the mechanism of checking and generating code coverage is to check the code coverage and if the overall coverage falls below 80% or the coverage of new files or new code falls below 80%, a comment with details will be uploaded. Additionally, comments are not added every time, but updated. If code coverage threshold is greater than 80% comment will be not uploaded. If the coverage of new code falls below 80%, the next step "Enforce minimum code coverage threshold on changed files" in CI will not pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -230,7 +230,6 @@ jobs:
             GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
             PR_NUMBER: ${{ steps.pr.outputs.pull_request_number }}
             GITHUB_REPOSITORY: ${{ github.repository }}
-            GITHUB_BASE_REF: ${{ github.base_ref }}
             WORKSPACE: ${{ github.workspace }}
         - name: "Enforce minimum code coverage threshold on changed files"
           if: ${{ steps.pr.outputs.pr_found == 'true' && steps.process_coverage.outputs.coverage-changed-files < 80.0 }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,6 @@ jobs:
     outputs:
       java_version: ${{ env.JAVA_VERSION }}
       java_version_supported_by_repo: ${{ env.JAVA_VERSION_SUPPORTED_BY_REPO }}
-      pr_number: ${{ github.event.pull_request.number }}
     steps:
       - uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v8.23.0
       - uses: actions/checkout@v4
@@ -136,8 +135,6 @@ jobs:
       matrix:
         subproject: [ "bulk-ingester", "common", "common-authentication", "common-test",
                       "live-ingester", "prediction-applier", "hxinsight-extension" ]
-    outputs:
-      pr_number: ${{ needs.pre_commit.outputs.pr_number }}
     steps:
       - uses: actions/checkout@v4
       - uses: Alfresco/alfresco-build-tools/.github/actions/free-hosted-runner-disk-space@v8.23.0
@@ -173,7 +170,6 @@ jobs:
       runs-on: ubuntu-latest
       needs:
         - test_application
-        - pre_commit
       permissions:
         pull-requests: write
       if: >
@@ -234,6 +230,7 @@ jobs:
             GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
             PR_NUMBER: ${{ steps.pr.outputs.pull_request_number }}
             GITHUB_REPOSITORY: ${{ github.repository }}
+            GITHUB_BASE_REF: ${{ github.base_ref }}
             WORKSPACE: ${{ github.workspace }}
         - name: "Enforce minimum code coverage threshold on changed files"
           if: ${{ steps.pr.outputs.pr_found == 'true' && steps.process_coverage.outputs.coverage-changed-files < 80.0 }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,10 +188,20 @@ jobs:
             fetch-depth: 0
         - name: Get PR number
           id: pr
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -200,23 +215,17 @@ jobs:
...
-            fail-emoji: ':red_circle:'
-            update-comment: true
-
+          id: process_coverage
+          run: |
+            pip install PyGithub
+            python ./scripts/ci/processCoverageReport.py
+          env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+            PR_NUMBER: ${{ steps.pr.outputs.pull_request_number }}
```
tpage-alfresco, line 224: Why don't we just use `github.event.pull_request.number` here? I don't really understand the flow for this value.

tpage-alfresco: Or reading more closely maybe we should use `$(gh pr view --json number -q .number || echo "")` here? I quick Google showed the PR number wasn't a standard env property :(

scripts/ci/processCoverageReport.py
```diff
@@ -0,0 +1,414 @@
...
+            print("Push event: Getting files changed in the last commit")
+            cmd = ['git', 'diff', '--name-only', 'HEAD^', 'HEAD']
+
+        print(f"Running command: {' '.join(cmd)}")
+        result = subprocess.run(cmd, capture_output=True, text=True)
+
+        if result.returncode != 0:
+            print(f"Error getting changed files with first approach: {result.stderr}")
+
+            cmd = ['git', 'diff', '--name-only', 'HEAD~10', 'HEAD']
```
tpage-alfresco, line 121: Do we often get an error trying to compare the base branch and the PR branch?

scripts/ci/processCoverageReport.py
```diff
@@ -0,0 +1,414 @@
...
+        return {
+            'overall': overall_coverage,
+            'files': file_coverage
+        }
+
+    except Exception as e:
+        print(f"Error parsing JaCoCo XML file {xml_path}: {str(e)}")
+        return {'overall': {}, 'files': {}}
+
+def get_git_changed_files():
```
tpage-alfresco, line 100: We already have a GHA that does this for the PMD scan. Can we reuse anything from that?

scripts/ci/processCoverageReport.py
```diff
@@ -0,0 +1,414 @@
...
+            files_coverage = report_coverage.get('files', {})
+
+            for coverage_path, coverage_info in files_coverage.items():
+                if changed_file.replace('\\', '/') in coverage_path:
+                    matches.append((changed_file, coverage_info))
+                    found = True
+                    print(f"Match found for {changed_file} in report #{report_index+1}")
+                    break
+
+                if short_name == os.path.basename(coverage_path).lower():
```
tpage-alfresco, line 167: I'm not sure I get this bit. Will this match files regardless of their location?

scripts/ci/processCoverageReport.py
```diff
@@ -0,0 +1,414 @@
...
+    print(f"Calculating overall {counter_type} coverage from {len(coverage_data)} reports...")
+    total_covered = 0
+    total_missed = 0
+
+    for data in coverage_data:
+        if 'overall' in data and counter_type in data['overall']:
+            total_covered += data['overall'][counter_type]['covered']
+            total_missed += data['overall'][counter_type]['missed']
+
+    if total_covered + total_missed > 0:
```
tpage-alfresco, line 199: Since `total_missed >= 0` then you can do:
```
overall_coverage = 0 if total_covered == 0 else (total_covered / (total_covered + total_missed)) * 100
```
I'd suggest just using total_lines everywhere though, because then this becomes:
```
overall_coverage = 0 if total_lines == 0 else 100 * total_covered / total_lines
```

.github/workflows/ci.yml
kamilbochno: I tried to use the standard `github.event.pull_request.number` but it doesn't work in this case because this step is only called in the 'push' action because that's how the tests are called so I have to take PR number from the step where the 'pull' action is called. I'll try to improve it a bit according to your suggestions.

scripts/ci/processCoverageReport.py
```diff
@@ -0,0 +1,414 @@
...
+                    missed = int(counter.get('missed', 0))
+
+                    file_coverage[file_path]['counters'][type_name]['covered'] += covered
+                    file_coverage[file_path]['counters'][type_name]['missed'] += missed
+
+        for file_path, file_data in file_coverage.items():
+            for counter_type, counter in file_data['counters'].items():
+                total = counter['covered'] + counter['missed']
+                counter['total'] = total
+                counter['percentage'] = (counter['covered'] / total * 100) if total > 0 else 0
```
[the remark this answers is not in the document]
kamilbochno: Removed

scripts/ci/processCoverageReport.py
```diff
@@ -0,0 +1,414 @@
...
+
+def get_env_variable(name, default=None, required=False):
+    """Get environment variable with logging and validation."""
+    value = os.environ.get(name, default)
+    if required and not value:
+        print(f"ERROR: Required environment variable '{name}' is not set")
+        return None
+
+    if value:
+        if 'TOKEN' in name or 'PASSWORD' in name:
```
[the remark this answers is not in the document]
kamilbochno: Added

scripts/ci/processCoverageReport.py
kamilbochno: I think it's unnecessary in this case so I removed it.

scripts/ci/processCoverageReport.py
tpage-alfresco: Ok - sounds good.

.github/workflows/ci.yml
kamilbochno: Now I added specifying PR by branch name and processing and adding comment to PR will run only if PR is found.

PR #994, opened 2025-06-24
[CIN-2503] Remove GHAS codeql & dependency graph

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-2503?focusedCommentId=2424722

After two months investigation of GHAS scans vs Veracode scans for hxinsight-connector, results shown more value in use of Veracode scans that we already use for this project, because of such output we've decided to stay with Veracode SAST and SCA scan + PMD and remove GHAS scanning tools


tpage-alfresco (approved):
LGTM - thanks Adam.

PR #1049, opened 2025-08-28
WIP: The unit test and integration test aren't done yet

tirthankar19:


bulk-ingester/src/main/java/org/alfresco/hxi_connector/bulk_ingester/processor/mapper/AlfrescoNodeMapper.java
```diff
@@ -62,21 +64,20 @@
 @Slf4j
 @Component
 @RequiredArgsConstructor
-public class AlfrescoNodeMapper
-{
+public class AlfrescoNodeMapper {
```
tpage-alfresco, line 67: Please can you use the formatter provided? You can use pre-commit to ensure you format everything automatically.

common/src/main/java/org/alfresco/hxi_connector/common/model/ingest/IngestEvent.java
```diff
@@ -62,4 +63,13 @@ public static class ContentInfo implements Serializable
...
         private String mimetype;
     }
+    @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+    @Data
+    @Accessors(fluent = true)
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class AncestorsInfo implements Serializable
+    {
+        private String parentId;
```
tpage-alfresco, line 73: Why do we need a class for this rather than just a string? Incidentally I'd prefer extracting this to its own class, which will make all the references to it neater.

PR #1363, opened 2026-04-09
CIN-11880 Remove requirement for Java 21 from Spring Boot components.

tpage-alfresco:


azakrzewski-hy (approved):

pom.xml
```diff
@@ -52,8 +52,8 @@
     </distributionManagement>
 
     <properties>
-        <java.version>21</java.version>
-        <alfresco-platform.java.version>${java.version}</alfresco-platform.java.version>
+        <java.version>17</java.version>
+        <alfresco-platform.java.version>21</alfresco-platform.java.version>
```
krdabrowski, line 56: Why `alfresco-platform.java.version` is specified to use Java 21? I think, that in CI actually we are passing there value: `11`

tpage-alfresco: Maybe I read this wrong - I thought this was the version for the JAMP, which by default will be built with ACS 26.1 now.

tpage-alfresco: Ok - I've split it into three versions total:
* java.version - used for Spring Boot components
* alfresco-platform.java.version - used with ACS 26.1 (default ACS)
* connector-common.java.version - used with both, and also with ACS 7.4

krdabrowski: > Maybe I read this wrong - I thought this was the version for the JAMP, which by default will be built with ACS 26.1 now.

This is correct, but the JAMP cannot be built using Java 21, because it won't work with ACS 7.4, 23 or 25

tpage-alfresco: Right - makes sense. We need to build the JAMP with Java 11, but ship it in a Docker image using ACS 26.1 and Java 21. Meanwhile the Spring Boot components will be Java 17 everywhere, and the common libraries will be built with Java 11. Thanks.

common/pom.xml
```diff
@@ -13,8 +13,8 @@
     <packaging>jar</packaging>
 
     <properties>
-        <maven.build.sourceVersion>${alfresco-platform.java.version}</maven.build.sourceVersion>
-        <maven.build.targetVersion>${alfresco-platform.java.version}</maven.build.targetVersion>
+        <maven.build.sourceVersion>${library.java.version}</maven.build.sourceVersion>
+        <maven.build.targetVersion>${library.java.version}</maven.build.targetVersion>
```
azakrzewski-hy, line 17: The common module is compiled with source/target Java 11, but it directly depends on spring-boot-starter-validation (Spring Boot 4.0.5). Spring Boot 4.x requires Java 17 as a minimum. While Maven might compile the source code at Java 11 level, the runtime dependencies themselves (Spring Boot 4) won't run on a Java 11 JVM.

krdabrowski (approved):

common/pom.xml
tpage-alfresco: The CI is testing this case, so I think we don't have an issue here.

PR #1385, opened 2026-04-27
[CIN-12434] Update flaky ATSTransformE2eTest by extending retry window

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-12434

The fix is to extend the retry window (60 attempts × 2000 ms ≈ 120 s) so the test rides out the transform pipeline's JMS cold start (currently ~15s total: 15 attempts × 1000 ms)


krdabrowski (approved):

azakrzewski-hy:
Change from https://github.com/Alfresco/hxinsight-connector/pull/1385/commits/738cb75ca44983247a9ee08d674329ebf36a52e4
In @BeforeAll, after the bucket is created and clients are wired up, the test now blocks until both transform engines log GET Transform Config version: — proof that ACS has successfully polled them and populated its TransformServiceRegistr

krdabrowski (approved):

PR #1388, opened 2026-04-30
[CIN-12602] Clear CWE-117 Improper Output Neutralization for Logs

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-12602

Adding sanitizer aligned with Veracode's official guidance:
https://community.veracode.com/s/article/How-to-Fix-CWE-117-Improper-Output-Neutralization-for-Logs

Utility for neutralizing untrusted data before it is written to log output, mitigating CWE-117 (Improper Output Neutralization for Logs / log forging)


common/src/main/java/org/alfresco/hxi_connector/common/util/LogSanitizer.java
```diff
@@ -0,0 +1,54 @@
...
+ * #L%
+ */
+package org.alfresco.hxi_connector.common.util;
+
+import static lombok.AccessLevel.PRIVATE;
+
+import lombok.NoArgsConstructor;
+
+/**
+ * Utility for neutralizing untrusted data before it is written to log output, mitigating CWE-117 (Improper Output Neutralization for Logs / log forging). Apply this at the trust boundary (where untrusted data first enters the application) so every downstream log statement is automatically safe.
```
tpage-alfresco: > Apply this at the trust boundary (where untrusted data first enters the application) so every downstream log statement is automatically safe.

Just a comment that we're not doing this in this PR. I don't actually think it makes sense to escape data just in case it's used in a log statement.

common/src/main/java/org/alfresco/hxi_connector/common/util/LogSanitizer.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+@NoArgsConstructor(access = PRIVATE)
+public final class LogSanitizer
```
tpage-alfresco: Is there no utility in a library for this? It seems like something I'd expect logging libraries to be able to handle on our behalf.

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/DiscoveryApiRepositoryInformation.java
```diff
@@ -86,7 +87,10 @@ private DiscoveryApiResponse getDiscoverApiResponse() throws IOException, Interr
         throwExceptionOnUnexpectedStatusCode(response.statusCode(), SC_OK);
 
         DiscoveryApiResponse discoveryApiResponse = objectMapper.readValue(response.body(), DiscoveryApiResponse.class);
-        log.trace("Discovery API response: {}", discoveryApiResponse);
+        if (log.isTraceEnabled())
+        {
+            log.trace("Discovery API response: {}", LogSanitizer.sanitize(discoveryApiResponse));
+        }
```
tpage-alfresco: More recently I've seen [fluent logging](https://www.slf4j.org/manual.html#fluent) with slf4j:
```
log.atTrace().log("Discovery...
```

common/src/main/java/org/alfresco/hxi_connector/common/util/LogSanitizer.java
azakrzewski-hy: Thank you Tom, you are right

common/src/main/java/org/alfresco/hxi_connector/common/util/LogSanitizer.java
azakrzewski-hy: No logging library ships a sanitizer that the static analyzer recognises:

- SLF4J/Logback offer %replace(%msg){'[\r\n]','_'} in pattern layout, but Veracode SAST can't see runtime logback config and won't clear the flaw.

- OWASP ESAPI has cleansers but is a heavy dependency for a 3-line need.

- Apache Commons Text StringEscapeUtils.escapeJava(...) is recognised by Veracode, but it's a new ~250 KB dependency we don't currently pull in, and it escapes Java-string-literal style (\n → \\n) which adds noise to logs.

We could switch to commons-text if you prefer it over current helper

common/src/main/java/org/alfresco/hxi_connector/common/util/LogSanitizer.java
azakrzewski-hy: Changed in: https://github.com/Alfresco/hxinsight-connector/pull/1388/commits/c7756f060fe5a6711c576874f6116c82388ae993

common-authentication/src/main/java/org/alfresco/hxi_connector/common/adapters/messaging/repository/DiscoveryApiRepositoryInformation.java
azakrzewski-hy: Implemented in: https://github.com/Alfresco/hxinsight-connector/pull/1388/commits/c7756f060fe5a6711c576874f6116c82388ae993

krdabrowski (approved):

common/src/main/java/org/alfresco/hxi_connector/common/util/LogSanitizer.java
tpage-alfresco: I really don't think this method of sanitising logs will scale to all the log lines in the application. It's going to be really hard to check all locations that the message might come from for every log call.

We should instead change the pattern (your first option), and then find out how to tell Veracode to stop flagging this as a warning. I think the pattern would be `%replace(%msg){'[\\x00-\\x1f\\x7f]', '_'}`

tpage-alfresco (changes requested):

common/src/main/java/org/alfresco/hxi_connector/common/util/LogSanitizer.java
azakrzewski-hy: Thank you Tom. Switched to the configuration-level approach. Added a single shared common/src/main/resources/logback-spring.xml that overrides Spring Boot's default CONSOLE_LOG_PATTERN (and FILE_LOG_PATTERN) to wrap the message conversion in %replace(%m){'[\x00-\x1f\x7f]', '_'}. Because common is a transitive dep of all four Spring Boot apps (bulk-ingester, live-ingester, nucleus-sync, prediction-applier), this single file protects every log statement in every app — no per-call-site changes required.

azakrzewski-hy: Implemented in https://github.com/Alfresco/hxinsight-connector/pull/1388/commits/f5b86073e6268f79a1dc6635c7399e07fbdec80d

common/src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tpage-alfresco, line 1: Will having this file here cause issues in the JAMP?

azakrzewski-hy: Yes, it would have. logback-spring.xml would have ended up at the root of the ACS classpath when the AMP is deployed inside alfresco.war.
Added a filter in hxinsight-extension/pom.xml that excludes logback-spring.xm
https://github.com/Alfresco/hxinsight-connector/pull/1388/commits/ded3e0b45482a2055002b4dbae7fcf36f6e97a69

tpage-alfresco (approved):
LGTM - thanks for taking the lead on this Adam.

PR #1399, opened 2026-05-08
ACS-11592 reliability audit & test suite

jakubkochman:
## Scope
ACS-11592 — adds an explicit reliability/observability layer to the live-ingester
and a chaos-style E2E suite that validates failure modes and recovery paths.
## Behavioural changes
- **Opt-in dead-letter wiring per route.** New `dead-letter-enabled` knob on
  `Repository.EventsSubscription`, `BulkIngester`, and `Transform.Response`,
  defaulting to `false`. No change to existing deployments — DLQ wiring is
  off until explicitly enabled.
- **DLQ + retry observability.** New Micrometer counters under
  `live_ingester_*` (DLQ totals, silent drops, unknown event-type drops,
  `@Retryable` attempts) tagged with the originating exception class.
  Exposed via `/actuator/metrics` (now permitted alongside
  `/actuator/health` in `AuthConfig`).
- **Bounded per-route HTTP timeouts.** `responseTimeoutMs` on
  `hyland-experience.ingester`, `hyland-experience.storage.location`,
  `hyland-experience.storage.upload`, `alfresco.repository`. Value `<= 0`
  means "leave Camel HTTP default" (no parameter emitted), matching
  `HttpFileUploader` / `AlfrescoRepositoryContentClient`.
- **Durable repo-event subscriptions.**
  `alfresco.repository.events-subscription.durable` + `.name` opt-in;
  `JmsClientIdConfigurer` sets the JMS clientId when enabled. URI value is
  URL-encoded before splicing. Multi-instance HA still not supported (single
  active subscriber).
- **Expanded retry policy.** `common/.../Retry.java` now retries
  `SocketException` (TCP RST surface uncovered by `S3TransientBlipReliabilityIT`).
- **Lock-free token reads.** `DefaultAccessTokenProvider` uses
  `ConcurrentHashMap` + per-provider `ReentrantLock`; concurrent fresh-token
  reads no longer serialize on a global monitor. Concurrent test added.
## New operator-facing config
| Property | Default | Notes |
|---|---|---|
| `alfresco.repository.events-subscription.durable` | `false` | enable durable JMS subscription |
| `alfresco.repository.events-subscription.name` | `LiveIngesterSubscription` | URL-encoded into URI |
| `alfresco.repository.events-subscription.dead-letter-enabled` | `false` | opt-in |
| `alfresco.repository.events-subscription.dead-letter-uri` | `activemq:queue:ActiveMQ.DLQ` | |
| `alfresco.repository.events-subscription.maximum-redeliveries` | `6` | |
| `alfresco.repository.events-subscription.redelivery-delay-ms` | `1000` | |
| `alfresco.repository.events-subscription.dead-letter-unsupported-types` | `false` | opt-in |
| `alfresco.repository.response-timeout-ms` | `0` | `<=0` → no per-request cap |
| `bulk-ingester.dead-letter-*` | analogous | |
| `transform.response.dead-letter-*` | analogous | |
| `transform.response.throw-failed-transforms` | `false` | opt-in DLQ for status=400 |
| `hyland-experience.ingester.response-timeout-ms` | `0` | |
| `hyland-experience.storage.location.response-timeout-ms` | `0` | |
| `hyland-experience.storage.upload.response-timeout-ms` | `0` | |
| `management.endpoints.web.exposure.include` | adds `metrics` | dashboards/prom scrape |
## Reliability test suite (new `reliability-tests` profile)
Toxiproxy + Testcontainers harness covering ACS / ActiveMQ / HXI / S3 /
transform / poison-pill / process-chaos failure modes
(`*ReliabilityIT.java`). Run with `mvn -pl e2e-test verify -Preliability-tests`;
regular e2es (`*E2ETest.java`) keep their existing path.
A new CI job `run_reliability_test` runs the suite on push events
(matrix: ACS 26.1.0 / Java 21, 40-min timeout, opt-out via
`[skip reliability]`). Not yet wired into release gating.
## Out of scope / follow-ups
- Multi-instance HA for durable subscriptions.


common-authentication/src/test/java/org/alfresco/hxi_connector/common/adapters/auth/DefaultAccessTokenProviderTest.java
```diff
@@ -121,4 +127,53 @@ void givenAuthenticationError_whenGetAccessToken_thenThrowException()
...
     }
 
+    /**
+     * Two concurrent {@code getAccessToken} calls that both observe the same expired cached token must result in a single auth-endpoint round-trip — the second caller, after acquiring the monitor, has to re-check the cache and pick up the token the first caller just refreshed.
+     *
+     * <p>
+     * Trigger: pre-seed an expired token, then race two threads. Thread A enters the {@code synchronized} block first; the mock {@code authenticate} blocks until Thread B has done its outside-the-monitor read of the cache map and is queued on the monitor. When the mock returns, Thread A puts the fresh token into the map and releases the monitor; Thread B acquires it. A correctly-implemented double-checked-locking pattern re-reads the map inside the monitor and short-circuits the refresh; the buggy pattern checks a stale local variable captured before the monitor and triggers a redundant {@code authenticate} call.
+     */
+    @Test
+    void shouldRefreshOnlyOnceWhenConcurrentCallersRaceForExpiredToken() throws Exception
```
jakubkochman, line 138: During testing I observed that `DefaultAccessTokenProvider` synchronizes incorrectly and if multiple threads detect stale token they would fire multiple http calls, instead of making one http call to retrieve another shareable token.

common-test/src/main/java/org/alfresco/hxi_connector/common/test/util/LogCaptureUtils.java
```diff
@@ -40,7 +40,7 @@
 import org.slf4j.LoggerFactory;
 
 @NoArgsConstructor(access = PRIVATE)
-public class LoggingUtils
```
jakubkochman, line 43: There were 2 classes named LoggingUtils. One is used in production for Camel messages, this one is essentially harness for asserting certain log messages. Changed the name for test one to remove ambiguity.

common/src/main/java/org/alfresco/hxi_connector/common/config/properties/Retry.java
```diff
@@ -78,7 +79,8 @@ public Retry(int attempts, int initialDelay, double delayMultiplier)
                 Stream.concat(RETRY_REASONS_BASIC.stream(), Stream.of(
                         HttpHostConnectException.class,
                         NoHttpResponseException.class,
-                        MalformedChunkCodingException.class)).collect(Collectors.toSet()));
+                        MalformedChunkCodingException.class,
+                        SocketException.class)).collect(Collectors.toSet()));
```
jakubkochman, line 81: The S3TransientBlipReliabilityIT discovered that a TCP reset would produce SocketException which was not retried as it was not contained in this list.

e2e-test/pom.xml
```diff
@@ -121,22 +136,44 @@
...
-                    </execution>
-                </executions>
                 <configuration>
-                    <includes>
+                    <includes combine.self="override">
                         <include>**/*E2eTest.java</include>
                         <include>**/*E2ETest.java</include>
                     </includes>
+                    <excludes>
+                        <exclude>**/*ReliabilityIT.java</exclude>
```
jakubkochman, line 163: Using another project as reference, the reliability tests are running exclusively on separate profile.

`mvn -pl e2e-test verify` -> E2eTest.java + *E2ETest.java only
`mvn -pl e2e-test verify -Preliability-tests` -> *ReliabilityIT.java only

Changes here made it possible to correctly inherit other properties.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/config/AuthConfig.java
```diff
@@ -52,7 +52,7 @@ public SecurityFilterChain securityFilterChain(HttpSecurity security) throws Exc
     {
         return security
                 .formLogin(AbstractHttpConfigurer::disable)
-                .authorizeHttpRequests(authorize -> authorize.requestMatchers("/actuator/health/**").permitAll())
+                .authorizeHttpRequests(authorize -> authorize.requestMatchers("/actuator/health/**", "/actuator/metrics/**").permitAll())
                 .authorizeHttpRequests(authorize -> authorize.anyRequest().denyAll())
```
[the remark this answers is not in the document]
jakubkochman: The same access seems to be present in elasticsearch-connector. If this is a real concern we can set a simple authorization in place or make the actuator visible only during tests.

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/util/DeadLetterChannelConfig.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * <p>
+ * {@link #deadLetterEnabled()} is the per-route on/off switch; route {@code configure()} methods are expected to gate the {@code errorHandler(DeadLetterChannels.forRoute(...))} call on it. The default of {@code true} preserves the always-on behaviour for routes whose dead-letter wiring is mandatory (repo-events, bulk-ingester) — i.e. any unhandled exception during in-delivery processing must land the message on the configured DLQ. Routes whose default is opt-in (transform-response) override the field on the record so deployments that don't want a DLQ entry per failed transform-response can keep the silent-drop semantics. With the DLC disabled the route falls back to Camel's {@code DefaultErrorHandler} + JMS-broker-side redelivery, which surfaces failures as broker-side redeliveries and ultimately to the broker's own dead-letter strategy rather than the route-scoped DLQ this helper builds.
+ */
```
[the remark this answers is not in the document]
jakubkochman: Adjusted default

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/repository/LiveIngesterEventHandler.java
```diff
@@ -73,4 +85,24 @@ public void configure()
...
+        {
+            log.info("Repository :: Subscribing to {} as a non-durable consumer. Events published while disconnected will be dropped by the broker. Set ALFRESCO_REPOSITORY_EVENTSSUBSCRIPTION_DURABLE=true to enable a durable subscription.", baseUri);
+            return baseUri;
+        }
+
+        String separator = baseUri.contains("?") ? "&" : "?";
+        String durableUri = baseUri
+                + separator
+                + "subscriptionDurable=true"
+                + "&durableSubscriptionName=" + subscription.name();
```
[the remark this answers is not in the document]
jakubkochman: Added encoding

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/hx_insight/storage/connector/PreSignedUrlRequester.java
```diff
@@ -77,7 +77,9 @@ public class PreSignedUrlRequester extends RouteBuilder implements StorageLocati
     public void configure()
     {
         // @formatter:off
-        String storageRequestEndpoint = integrationProperties.hylandExperience().storage().location().endpoint() + ApplicationInfoProvider.USER_AGENT_PARAM;
+        String storageRequestEndpoint = integrationProperties.hylandExperience().storage().location().endpoint()
+                + "&httpClient.responseTimeout=" + integrationProperties.hylandExperience().storage().location().responseTimeoutMs()
+                + ApplicationInfoProvider.USER_AGENT_PARAM;
```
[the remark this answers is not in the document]
jakubkochman: Adjusted setting timeout to match

live-ingester/src/main/java/org/alfresco/hxi_connector/live_ingester/adapters/messaging/hx_insight/HxInsightEventPublisher.java
```diff
@@ -73,7 +73,9 @@ public class HxInsightEventPublisher extends RouteBuilder implements IngestionEn
     public void configure()
     {
         // @formatter:off
-        String ingestionEndpoint = integrationProperties.hylandExperience().ingester().endpoint() + ApplicationInfoProvider.USER_AGENT_PARAM;
+        String ingestionEndpoint = integrationProperties.hylandExperience().ingester().endpoint()
+                + "&httpClient.responseTimeout=" + integrationProperties.hylandExperience().ingester().responseTimeoutMs()
+                + ApplicationInfoProvider.USER_AGENT_PARAM;
```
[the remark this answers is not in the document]
jakubkochman: Done the same way

damianujma (approved):
Great work! 🚀🚀
Can we set configuration for a reliable system (e.g. durable subscription) by default?

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/active_mq/ActiveMqBacklogDrainReliabilityIT.java
```diff
@@ -0,0 +1,126 @@
...
+ * <p>
+ * Gated by the {@code reliability-tests} profile; opt-in with {@code mvn -pl e2e-test -am verify -Preliability-tests -Dit.test=ActiveMqBacklogDrainReliabilityIT}.
+ */
+@Slf4j
+@SuppressWarnings({"PMD.FieldNamingConventions", "PMD.TestClassWithoutTestCases"})
+public class ActiveMqBacklogDrainReliabilityIT extends BaseReliabilityIT
+{
+    private static final String PARENT_ID = "-my-";
+    private static final String REPO_EVENT_TOPIC = "alfresco.repo.event2";
+    private static final int BACKLOG_SIZE = 3;
```
kushal-banik-hyland, line 53: Any specific reason we are setting up this backlog size small batch? Considering ActiveMQ down for a certain time there could be lot of events so should we also test that or any specific number after that number of requests starts dropping from the actual consumed

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/active_mq/ActiveMqBandwidthCutReliabilityIT.java
```diff
@@ -0,0 +1,144 @@
...
+    private static final Duration INITIAL_DELAY = Duration.ofSeconds(1);
+    private static final Duration BANDWIDTH_CUT_DURATION = Duration.ofSeconds(5);
+    private static final int EVENTS_DURING_CUT = 3;
+    private static final long PUBLISH_GAP_MS = 800L;
+    private static final long SETTLE_AFTER_RECOVERY_MS = 2_000L;
+    private static final int CONVERGENCE_DELAY_MS = 1_000;
+    /**
+     * At-least-once is allowed; uncontrolled redelivery is not. The connector's reconnect path can produce a small number of duplicates after the broker decides the consumer is gone, but tens of duplicates would indicate a redelivery loop.
+     */
+    private static final int MAX_DUPLICATES_PER_NODE = 5;
```
kushal-banik-hyland, line 61: Any specific reason we are taking max allowed as 5. As per linux the clipping is done at TCP_RTO_MAX which is 120 seconds. 
Sequence will be like
Retry 1: 1 second (RTO = 1s)
Retry 2: 3 seconds (RTO doubles to 2s)
Retry 3: 7 seconds (RTO doubles to 4s)
Retry 4: 15 seconds (RTO doubles to 8s)
Retry 5: 31 seconds (RTO doubles to 16s)
And so on. So There could be 2-3 max duplicates. Idk if I have miscalculated something!

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/active_mq/ActiveMqFlappingReliabilityIT.java
```diff
@@ -0,0 +1,118 @@
...
+                CONVERGENCE_DELAY_MS);
+
+        ToxicPlanner<eu.rekawek.toxiproxy.Proxy> planner = new ToxicPlanner<>(
+                environment().activemqProxy(),
+                ToxicPlans.disableAndEnableProxyContinuously());
+        log.info("[reliability] Starting flapping proxy for ~{}", FLAPPING_DURATION);
+        planner.start();
+        try
+        {
+            Thread.sleep(FLAPPING_DURATION.toMillis());
```
kushal-banik-hyland, line 75: During tis period what if we are publishing huge number of messages? would it lost or we will have multiple duplicates? Can we test that.

We also have explicitly tested some like sending RST instead of FIN and also introducing jitter but this is also an interesting case where we also can see what happens during this time

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/active_mq/ActiveMqBacklogDrainReliabilityIT.java
jakubkochman: I tried to make tests isolated when it comes to introduced chaoses. Here we only cut connection between connector and broker, so events that arrive during chaos should be preserved. This test was enough to spot a problem with default configuration that was not setting durable subscription. The broker detected that consumer disconnected and was simply dropping the ACS events as nobody was listening for them.

Anyway the idea to exercise amount is good, I will try to create some tests to explore it.

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/poison_pill/RepoEventPoisonPillReliabilityIT.java
```diff
@@ -0,0 +1,98 @@
...
+
+        RetryUtils.assertWithRetry(() -> {
+            assertThat(WiremockCounts.ingestionEventsFor(sentinelNode.id()))
+                    .as("liveness: sentinel event published after the poison pill must reach HX Insight — failure here means the poison pill stopped the route")
+                    .isGreaterThanOrEqualTo(1);
+            assertThat(environment().jolokia().dlqDepth())
+                    .as("no silent drop: malformed message must land on ActiveMQ.DLQ — a zero here means the route ACK'd it and the broker dropped it on the floor")
+                    .isGreaterThanOrEqualTo(1);
+            assertThat(environment().jolokia().dlqDepth())
+                    .as("no retry storm: poison pill should be parked exactly once, not redelivered")
```
kushal-banik-hyland, line 78: So it will not be retried right? Will it stay here or dropped after certain time

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/active_mq/ActiveMqRestartReliabilityIT.java
```diff
@@ -0,0 +1,182 @@
...
+            List<Node> backlog = new ArrayList<>(BACKLOG_SIZE);
+            for (int i = 0; i < BACKLOG_SIZE; i++)
+            {
+                backlog.add(createNode("amq-backlog-" + i + ".txt", "Backlog event #" + i));
+            }
+
+            log.info("[chaos] re-enabling Toxiproxy — slow Wiremock responses keep the drain in progress while we restart the broker");
+            try
+            {
+                environment().activemqProxy().enable();
```
kushal-banik-hyland, line 138: If we are making the proxy enable before the graceful stop then those events will come to us usual but ig our intension to test here how it persists those unsubmitted events in ActiveMQ after graceful stop and restart?

kushal-banik-hyland (approved):
Great!

SatyamSah5 (approved):

mousumi-podder (approved):

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/acs/AcsSlowDuringAmqDisconnectReliabilityIT.java
```diff
@@ -0,0 +1,185 @@
...
+                Node chaosNode = createSentinel("composed-acs-amq-chaos-" + i + ".txt", ("chaos-" + i).getBytes());
+                chaosNodes.add(chaosNode);
+                log.info("[reliability] Chaos event {} / {}: node id {}", i + 1, CHAOS_NODE_COUNT, chaosNode.id());
+            }
+            Thread.sleep(AMQ_DISCONNECT_DURATION_MS);
+        }
+        finally
+        {
+            log.info("[reliability] Re-enabling Toxiproxy — backlog drain begins through still-slow ACS");
+            if (!environment().activemqProxy().isEnabled())
```
kushal-banik-hyland: Any chance of concurrency or something as we are checking if some other worker has already enabled this proxy or not?

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/throughput/NoChaosBaselineReliabilityIT.java
```diff
@@ -0,0 +1,265 @@
...
+            List<Future<?>> futures = new ArrayList<>(count);
+            for (int i = 0; i < count; i++)
+            {
+                int idx = i;
+                futures.add(pool.submit(() -> {
+                    try (InputStream content = new ByteArrayInputStream((label + "-" + idx).getBytes()))
+                    {
+                        Node node = environment().repositoryClient()
+                                .createNodeWithContent(PARENT_ID, "baseline-" + label + "-" + idx + ".txt", content, "text/plain");
+                        int done = completed.incrementAndGet();
```
kushal-banik-hyland, line 231: Any Possibility we can use `nextThreadId` here? Are those same?

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/poison_pill/RepoEventPoisonPillReliabilityIT.java
jakubkochman: With default settings whatever lands in DLQ stays there forever. You can configure some cleanup strategies or replay messages (like if the reason was external and it's safe to do).

The idea for this test in general is to observe and document what happens when we push a broken event into the queue. In short:
- We want to log that we received something unexpected
- We don't want to block our processing by ACK the event and not throwing exceptions that would interrupt it
- We don't want to loose details about the event by storing it in DLQ and allow to replay it
- We want to retry, but to some limit

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/active_mq/ActiveMqRestartReliabilityIT.java
jakubkochman: Initially this test was enabling the toxic network for too short period of time so I couldn't actually verify that application noticed that something was wrong. It was possible to consume all events and then it would just loose connection to the queue. 

So the idea is to make downstream live-ingesting api slow and effectively blocking, making the events in the queue accumulate. Then we interrupt connection with activemq:
- The events should remain on the queue even if connector disconnected. I know it depends on activemq implementation, but here we can at least verify the connection settings.
- The connector should re-establish connection and continue consuming events.

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/acs/AcsSlowDuringAmqDisconnectReliabilityIT.java
jakubkochman: The whole reliability suite runs sequentially. The only thing shared could be test context or docker containers if given test is sharable: 
* it does not require a different configuration or 
* the introduced chaos is easy to cleanup

I think I can drop the `if` as Toxiproxy should handle it as idempotent operation.

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/throughput/NoChaosBaselineReliabilityIT.java
jakubkochman: I don't think I understand the question. You mean `completed` and `nextThreadId`? These are different things.

`nextThreadId` is just id of worker that is submitting nodes
`completed` is used to periodically log progress of submitting events

kushal-banik-hyland: Correct! I was thinking in the direction like those two must be same (assuming every task is given to every thread).
Ya my bad. Nothing to change

e2e-test/src/test/java/org/alfresco/hxi_connector/e2e_test/reliability/active_mq/ActiveMqRestartReliabilityIT.java
kushal-banik-hyland: Yes. `KahaDb` is responsible for doing that. https://activemq.apache.org/components/classic/documentation/kahadb
So it will retry with unpublished events again

jakubkochman:
Closing in favor of https://github.com/Alfresco/alfresco-cic-connector/pull/62

PR #1441, opened 2026-06-23
[CIN-14309] Upgrade Spring-Boot from 4.0.6 to 4.1.0

azakrzewski-hy:
https://hyland.atlassian.net/browse/CIN-14309

**Changes in the PR**

- Spring Boot 4.0.6 → 4.1.0 (pom.xml).
- ActiveMQ client conflict — Boot 4.1's spring-boot-starter-activemq now pulls activemq-client:6.2.6, which clashed with the activemq-client-jakarta:5.19.x from camel-activemq (both ship the same org.apache.activemq.* packages, corrupting the OpenWire wire format → Transport scheme NOT recognized: [vm] / EOFException). Excluded the older jakarta client in bulk-ingester, live-ingester, and prediction-applier, keeping the single Boot-managed 6.2.x client (OpenWire-compatible with the 5.19 broker).
- WireMock test container — disabled gzip (--disable-gzip) so empty 202 responses aren't returned with Content-Encoding: gzip, which made Camel's HTTP client fail decompressing an empty body (EOFException).
- ActiveMQ test container — removed the jetty.xml overlay that broke the 5.19 broker startup (stopped the OpenWire connector → connection refused), and hardened the readiness check to wait for both the broker "started" log and the OpenWire port 61616 to be listening. Deleted the now-unused jetty.xml resource.
- live-ingester surefire — added combine.children="append" to the module's test excludes so the parent POM's integration/E2E exclusions are preserved and those container-backed tests no longer run during mvn test.
- Updated DockerContainersTest assertions to match the new WireMock options.


damianujma (approved):
LGTM

